PR #18, opened 2018-02-13
Profile image cropping

tristan-orourke:


tristan-orourke:
Dropping file creates two thumbnails for some reason

PR #42, opened 2018-02-28
G2 bugfixes

gggrant:


gggrant:
Tristan, can you make sure your 'at' label entry was merged correctly?

PR #102, opened 2018-05-03
G2 mergefix

gggrant:


gggrant:
Added changes lost in a previous merge

PR #164, opened 2018-06-06
Qf update base content 20180528

qfung:
Please review my code and merge if possible.


qfung:
NOTE: Application Preview localization is not complete, that will be completed with the next round of translations received from project team.

PR #297, opened 2018-08-24
Reverting messy merges of gb_postres into dev

tristan-orourke:


tristan-orourke:
This basically reverts dev to before gb_postgres was partially merged in. Things were merged, reverted, and merged again in a way that deleted feature/job.show_controller. This restores that feature. Merging gb_postres into dev when its done will have to be done very carefully.

PR #304, opened 2018-08-28
Gb postgres

gggrant:


tristan-orourke:
This won't work for Production though. Right now the whole database is lost every time our docker server shuts down.

PR #312, opened 2018-09-05
Add ability to control manager path from env file

tristan-orourke:


joshdrink:
Tested both applicant and manager logins. Works!

PR #316, opened 2018-09-06
Downgrade from Laravel 5.6 to 5.5

tristan-orourke:


tristan-orourke:
I've tested this on both portals, seems to be working

PR #361, opened 2018-10-15
Users can't edit submitted applications, but can still view preview

tristan-orourke:


gobyrne:
Ok, it was breaking because a job got seeded that had no essential criteria, which broke things. I can still edit after job submission though. For example, the open ended questions on page one can be changed after a job is "submitted". Also, the signature and date do not appear when you look at your previously submitted job.

gobyrne:
Ok, working better now. I had my environment set to "force_admin=true"

Issue #439, opened 2018-10-25
Applicant- Design & translate form for password recovery


gobyrne:
# Description
The form where you give a new password after recovery needs to be styled and translated.

# Required for Completion
Add/remove others as necessary for this specific task.
- [ ] Run Lighthouse (Chrome Dev Tools)
- [ ] Run WAVE Evaluation
- [ ] Browser Compatibility Check
- [ ] Update Translation Spreadsheet
- [ ] Does this affect configuration? (If so discuss with team)
- [ ] Create pull request to `dev`
- [ ] Address issues found during peer review
- [ ] Have someone else accept pull request into `dev`

gobyrne:
@joshdrink Just wanted to flag this to you. Probably higher priority than mobile fixes.

PR #440, opened 2018-10-25
Issue #439 Styles Reset Password UI to match Auth

joshdrink:


gobyrne:
Ok, let's handle that seperately and merge this one in.

PR #461, opened 2018-10-31
Validateaboutme

Djama613:
Security/Validation rules applied to forms in the About Me section under MyProfile.


Djama613:
Updated to check its an actual linkedin url.

PR #482, opened 2018-11-15
Show days remaining instead of weeks or months

tristan-orourke:


jprembis:
This was also bothering me. What about the following solution?

    public function timeRemaining($str_pass, $str_fail) {
        if ($this->close_date_time->isFuture()) {
            $time = $this->close_date_time->diffForHumans(null, true);
            return str_replace_first(':time', $time, $str_pass);
        } else {
            return $str_fail;
        }
    }

Then, in the job_index view for example, we would simply have

    {{ job.timeRemaining(job_index.index.time_remaining,
                         job_index.index.job_closed) }}

If you want, I could submit a PR.

PR #492, opened 2018-11-19
Bugfix/greencheckmark

Djama613:
small fix

resolves greencheckmark on need to have and nice to have skills when nothing is declared

Will be updated when Ajax is implemented.


tristan-orourke:
This will be merged into the forthcoming AJAX forms pull request.

PR #508, opened 2018-11-26
Database data persists on container restart

gggrant:
To test:

I renamed the DB container to ‘Postgres’, requires a change in .env:
DB_HOST=postgres

Create the DB data volume in Docker:
docker volume create pgdata

NOTE:
Do not use docker-compose down with -v or the data volume will be erased

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -11,7 +11,7 @@ FORCE_ADMIN=false
 DEBUGBAR_ENABLED=false
 
 DB_CONNECTION=pgsql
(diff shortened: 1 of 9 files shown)


gggrant:
To test:

I renamed the DB container to ‘Postgres’, requires a change in .env:
DB_HOST=postgres

Create the DB data volume in Docker:
docker volume create pgdata 

Do not use docker-compose down with -v or the data volume will be erased

PR #512, opened 2018-11-29
Added application review panel for managers

tristan-orourke:


gggrant:
Tested job sorting and adding notes

PR #515, opened 2018-12-06
Skill Declarations submit with ajax

tristan-orourke:
This includes server-side verification, and ux works on profile AND on applications, which includes a save all button.


gggrant:
Every time you add a new skill on My Profile, the textarea gets another red asterisk.

tristan-orourke:
This requires you to run `npm install` to function on IE

PR #525, opened 2018-12-10
Make login email not case sensitive

tristan-orourke:


Djama613:
Working to properly implement this bugfix. Found a simple/handy guide that doesn't apply to our problem/solution but remains a good resource for similar issues in the future.

https://tech.fleka.me/translate-laravels-reset-password-email-b0f1d6e4709a

PR #526, opened 2018-12-11
Release/PHP 7.1

gggrant:
Requires updating Composer and rebuilding Docker containers via CLI


gggrant:
Closing, see release/php7.2

Issue #555, opened 2018-12-21
[Design] - Redesign "Education" & "Equivalent Experience"


JasmitaChawla:
# Design Goal
**Please describe the ask.**

Add a definition for _My Lived/Equivalent Experience_. Under _My Profile_, it says _My Equivalent Experience_, but under _My Applications,_ it says _My Lived Experience_, change this to _My Equivalent Experience_. 

Note: The text has to be reviewed. 

**Related Features:** 
- Applicant Timeline
- Applicant Profile

**Related Issues:**
- #1435

# Wireframes

## Current UI
<img width="1406" alt="Screen Shot 2019-03-11 at 10 19 54 AM" src="https://user-images.githubusercontent.com/16977254/54131000-d87ba900-43e7-11e9-9b5f-ae051bedd895.png">

## Proposed UI
<img width="735" alt="definition for my equivalent/lived experience" src="https://user-images.githubusercontent.com/38079220/50351280-b15fba80-050f-11e9-901e-f5f40014af77.png">

gobyrne:
I standardised the text in Cleanup web.php merge request #571.

Jerryescandon:
@gobyrne  what are the rules for an equivalent experience to be accepted by a manager?
 - minimum number of years?
 - related field?

PR #572, opened 2019-01-04
Gobyrne text cleanup

gobyrne:
Small changes to lang files and deleting unused file


gobyrne:
This addresses the inconsistent language flagged in #555

PR #575, opened 2019-01-08
Add project defaults and update phpcbf path to be cross-platform

Khristophor:


Khristophor:
@tristan-orourke have you `composer update`'d recently? I added the extra sniff through that.

PR #599, opened 2019-01-09
Fix bug that blocked ajax delete from working

tristan-orourke:


resources/views/common/skill.html.twig
```diff
@@ -83,7 +83,7 @@
         modal-target-object"
 
     data-item-saved="{{ skill_declaration ? "true" : "" }}"
-    data-item-url="{{ route("skill_declarations.create") }}"
+    data-item-url="{{ route("skill_declarations.destroy", ":id") }}"
```
Khristophor: Just curious about the route change here.

tristan-orourke: This just makes it consistent with other instances of _data-item-url_, which all use the objects delete route for this url template.

PR #600, opened 2019-01-10
Sort experience by date

Khristophor:


app/Http/Controllers/ExperienceController.php
```diff
@@ -30,12 +30,27 @@ public function show(Applicant $applicant)
...
-     * @param  \App\Models\Applicant  $applicant
-     * @return \Illuminate\Http\Response
+     * @param \Illuminate\Http\Request $request   Incoming request object.
+     * @param \App\Models\Applicant    $applicant Incoming applicant object.
+     *
+     * @return \Illuminate\View\View|\Illuminate\Contracts\View\Factory
      */
     public function edit(Request $request, Applicant $applicant)
     {
+        $applicant->load(
```
tristan-orourke: This works for the Applicant Profile, but leaves experience unsorted on other pages, such as step 2 of the application, and Application Preview. Would it be possible to make this part of the Applicant object itself?

Khristophor: I can take a look at that.

PR #631, opened 2019-01-15
Unique Email Validation

tristan-orourke:


tests/Unit/Validators/UpdateApplicationProfileValidatorTest.php
```diff
@@ -0,0 +1,173 @@
...
+            'tagline' => 'GO FOR IT',
+        ];
+        $this->be($applicant->user);
+        $validator = new UpdateApplicationProfileValidator($applicant);
+        $this->assertFalse($validator->isValid($data));
+    }
+
+    public function testUpdatePasswordFailsWithBadOldPassword()
+    {
+        $applicant = factory(Applicant::class)->create();
```
Khristophor: I _think_ an error was thrown because the factory doesn't use unique email addresses (necessarily). Should be able to resolve this by updating the factory with `->unique()` on the email address.

You can also add a `setUp()` method and create a couple applicants there instead of a new one in every test, saves a bit of time accessing the DB. See the JobControllerTest for an example.

Khristophor: Hmm my bad, it's set to unique already. Not sure why it's erroring.

Issue #606, opened 2019-01-10
Applicant - Accordion triggers not obviously clickable


gggrant:
# Description
Accordion triggers do not register as a link, both on profile and job application. Users with assistive technology would not know that they can be clicked on to perform addition actions.

@joshdrink would it be much effort to change this without impacting the UI?

# Images
<img width="1161" alt="screen shot 2019-01-10 at 1 49 21 pm" src="https://user-images.githubusercontent.com/31678168/50989958-ca400800-14de-11e9-8326-8b298a08a98d.png">

# Required for Completion
Add/remove others as necessary for this specific task.
- [ ] Run Lighthouse (Chrome Dev Tools)
- [ ] Run WAVE Evaluation
- [ ] Browser Compatibility Check
- [ ] Update Translation Spreadsheet
- [ ] Does this affect configuration? (If so discuss with team)
- [ ] Create pull request to `dev`
- [ ] Address issues found during peer review
- [ ] Have someone else accept pull request into `dev`

joshdrink:
Somehow these aren't button elements?!?! Anyway, easy enough to fix - I'll make it similar to the design system until that's implemented.

PR #660, opened 2019-01-25
Changes Accordion Triggers to Buttons #606

joshdrink:
In order to address #606, accordion trigger objects have been changed to `button` elements with `type="button"`. @gggrant Let me know if this fixes them in the way you intended.


joshdrink:
@gggrant Done, @gobyrne `TRANSLATION NEEDED` :')

joshdrink:
Fixed a bug where the accordions were being triggered twice on "enter".

Issue #610, opened 2019-01-10
Applicant - Elements using class="hidden" ignored by screen readers


gggrant:
# Description
The styles used for the "hidden" class are being ignored by screen readers, but are intended to be read by screen readers as a visual alternative. Elements include Step 1-5 links, delete buttons that use an X, profile pictures.

See this link for suggested styles for hiding content as well as allowing for SR interpretation:
https://webaim.org/techniques/css/invisiblecontent/

Create a separate class for SR-only content if we are hiding content in other places for different reasons with "hidden".

# Required for Completion
Add/remove others as necessary for this specific task.
- [ ] Run Lighthouse (Chrome Dev Tools)
- [ ] Run WAVE Evaluation
- [ ] Browser Compatibility Check
- [ ] Update Translation Spreadsheet
- [ ] Does this affect configuration? (If so discuss with team)
- [ ] Create pull request to `dev`
- [ ] Address issues found during peer review
- [ ] Have someone else accept pull request into `dev`

joshdrink:
So I built the hidden class to deliberately be ignored by screen readers. I'll take a look at the elements you're describing, but this is intended behaviour.

PR #662, opened 2019-01-25
WIP Bug/hidden accessibility

joshdrink:
I've added a few things to address the primary requests in issue #610:
- There is now a `.invisible` class that hides an element visually but not from screen readers.
- step links, delete buttons, and applicant facing profile photos have all been updated with the invisible class and given an `aria-label` attribute to boot.

@gggrant Do you mind reviewing these changes for me?


joshdrink:
@gggrant Done!

Issue #637, opened 2019-01-17
All browsers - French skill status not displaying correctly


gobyrne:
# Description
The text "status: competenseAjoute" is not correct in the screenshot below.

# Images
![image](https://user-images.githubusercontent.com/2494551/51349193-a9ddf380-1a72-11e9-85d5-de25b79482f8.png)

gobyrne:
The interaction is a little awkward in general, so at some point we'll need a design update on the skills. For now it's OK though. FYI, I also added a space.

PR #663, opened 2019-01-25
WIP Fixes Skill Claim in French #637

joshdrink:
Fixes the "Status: Claimed" area of a skill when translated in French. @gobyrne Please review to ensure this is now acting as you expect.


gobyrne:
Fixes #637

PR #672, opened 2019-01-29
Validated composer.json

gggrant:
Add license, remove exact version constraints, use caret instead of tilde operator.

https://stackoverflow.com/a/43145865


gggrant:
We don't have such a convention. `^1.3.2` should update to `1.4` if it doesmn't break constraints.I think this is a more accurate explanation of the operator:
```
    // ~ | allows last digit specified to go up
    "vendor/package": "~1.3.2", // >=1.3.2 <1.4.0
    "vendor/package": "~1.3", // >=1.3.0 <2.0.0

    // ^ | doesn't allow breaking changes (major version fixed - following semver)
    "vendor/package": "^1.3.2", // >=1.3.2 <2.0.0
    "vendor/package": "^0.3.2", // >=0.3.2 <0.4.0 // except if major version is 0
```

https://getcomposer.org/doc/articles/versions.md#next-significant-release-operators

Issue #655, opened 2019-01-24
Profile Validation fail dumps all edited field data


xander-miller:
# Description
When a user is filling out narrative questions `https://talent.local.ca/en/profile/1/about` if a validation fails when updating data in all fields that were edited is lost. 

Ideally when a validation fails on a long form submission form data data is preserved so it doesn't have to be filled out again. 

# Steps to Reproduce
1. Go to 'https://talent.local.ca/en/profile/1/about'
2. Enter a random string into the linkedIn Profile field.
3. Put a bunch of text into one of the text area answer fields.
4. Click Save Button
5. Watch your answer text disappear.

gggrant:
Page has changed.

PR #673, opened 2019-01-29
WIP Validation for Create Job Poster form

gggrant:
Errors display, validation rules for most fields.

To do: Populate fields when validation fails, improve rules, improve display of error messages, language of error messages.


app/Http/Controllers/JobController.php
```diff
@@ -313,6 +313,35 @@ function ($skillLevel) use ($skillLangs) {
      */
     public function store(Request $request, JobPoster $jobPoster = null)
     {
+        // Job Poster validation rules
+        $validatedData = $request->validate([
```
tristan-orourke: This validation needs to run only when the poster is being submitted. You should be able to save an incomplete draft.

gggrant: Only runs on publish now. Could use your help moving the rules to JobPosterValidator or someone else more appropriate.

gggrant:
Closing for now, depends on #655

Issue #643, opened 2019-01-18
Admin - Look into PHPCS and/or Scrutinizer for linting lang files


gggrant:
# Required for Completion
Add/remove others as necessary for this specific task.
- [ ] Run Lighthouse (Chrome Dev Tools)
- [ ] Create pull request to `dev`
- [ ] Address issues found during peer review
- [ ] Have someone else accept pull request into `dev`

gggrant:
Similar to #419

PR #679, opened 2019-01-30
Gobyrne templates 2

gobyrne:
Addresses updates to definition of done discussed as part of #643


gobyrne:
@joshdrink did I do something horribly wrong with the templates? Not sure why the tests wouldn't pass

PR #680, opened 2019-01-31
Adds Application Step 6

joshdrink:
@tristan-orourke The template for this exists now - I've also modified the 5th step and added some logic for the 6th step in the tracker. You'll want to uncomment the tracker partial in the 6th step's template too! Shout if you need me for anything beyond that :)

@gggrant We'll want your eyes on this after Tristan does his thing.

@gobyrne I think there's one or two simple TRANSLATION NEEDED spots too!


gobyrne:
Translations done

Issue #419, opened 2018-10-23
Admin - Create Unit Tests for Translation


Djama613:
Resolved by #683 

Many issues found by tests, resolving before merge.

# Description
Unit Test for Translation (Pass Criteria)

- French file exists
- French Key/Value exists
- French Value =/= English Value (Would be nice if it offered a warning without stopping the test optional)
- Word "translate" does not exist
- Loops over all English Lang files.
- Then Loop over all French Files (get rid of unused French Test, NOTE: This might be revised as there's some text that will need to be addressed by other team members first.)

# Images

![image](https://user-images.githubusercontent.com/43754271/47385607-85bf8000-d6d8-11e8-87c6-fb47773253ef.png)

PR #683, opened 2019-01-31
Tests for validating lang files. Run tests with --verbose for details

tristan-orourke:
Issue #419
Run `docker-compose exec talentcloud sh -c "vendor/bin/phpunit --verbose --filter LangFilesTest"` for useful details on missing lang entries. Note that these tests aren't passing yet.


gggrant:
Resolves #419

PR #706, opened 2019-02-07
WIP Composer integrated with Docker

gggrant:
Review options for composer:
- Composer binaries in Dockerfile, allowing commands in the container with `docker exec ...`
- Composer image builds in docker-compose.yml and runs install


xander-miller:
Just tested it in linux. Doesn't work without php natively installed.

gggrant:
These improvements will be part of #722

PR #710, opened 2019-02-13
Admin Portal

tristan-orourke:
Starts with ability to edit Job Posters, edit Users, and edit/add Skills.


tristan-orourke:
New views were added in `resources\views\errors` that change how 403/404/etc errors appear. Are we happy with that?

PR #732, opened 2019-02-21
Task/manager publish job policy

### LINKED ISSUE #729: Manager - Manager's 'update job' permission should not extend to publishing jobs
# Description
Publishing a job poster should require a separate permission than creating or updating it. This publish permission should belong to Admin but not Manager roles.

xander-miller:
Resolves #729 
Managers should not be able to update published jobs

* Adds policy to stop managers from updating published jobs.
* Adds Unit Policy Test and Feature Test to verify

### DIFF
--- a/app/Policies/JobPolicy.php
+++ b/app/Policies/JobPolicy.php
@@ -48,9 +48,10 @@ class JobPolicy extends BasePolicy
      */
     public function update(User $user, JobPoster $jobPoster)
     {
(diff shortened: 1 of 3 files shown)


gobyrne:
Works, but it also removes the ability for Admins to edit job posters. Discussed this with @xander-miller. As a temporary solution, he will add this functionality back for admins, but we will also start a conversation about creating new routes for /admin

Issue #768, opened 2019-03-04
Manager - Styling is broken for Review panel when viewing application


tristan-orourke:
# Description
As a manager, when viewing an application, the review panel isn't styled correctly.

# Steps to Reproduce

# Images
Currently appears like this:
![image](https://user-images.githubusercontent.com/4313717/53763218-2c423b80-3e98-11e9-823d-a48e1994f8fc.png)

It should appear like this:
![image](https://user-images.githubusercontent.com/4313717/53763241-3d8b4800-3e98-11e9-9fcf-91e4f24be73c.png)

PR #778, opened 2019-03-05
Fixes the applicant review UI when a manager is viewing their applica…

joshdrink:
…tion, as well as some minor UI alignment.

Addresses Issue #768


resources/assets/js/ziggy.js
```diff
@@ -1,8 +1,8 @@
...
-        baseUrl: 'https://talent.local.ca/',
```
tristan-orourke: We should set this back to `talent.local` before merging into dev.

PR #792, opened 2019-03-09
Slacktee in builds (sends Lighthouse crawl results to Slack)

gggrant:
Can be used for other output/logs as well.


.travis.yml
```diff
@@ -80,6 +77,8 @@ after_script:
 # Slack notifications / do not send emails
 notifications:
   email: never
-  slack: talent-cloud:ASWQRTtW3oT2C2DHYaBzSCJg
-  on_success: always
-  on_failure: change
+  slack:
+    rooms:
+      - talent-cloud:ASWQRTtW3oT2C2DHYaBzSCJg
```
Khristophor: Is this an access token for Slack posting? Any concerns around having this checked in?

gggrant: I've hidden it with other services. Wasn't sure. I'll add an encrypted env variable on Travis...

Issue #806, opened 2019-03-20
Hiring Manager - Application Screening Navigation


joshdrink:
# Description
Lauren's Comment: 
```navigation for screening applications on the new portal needs work. Especially need a “back” button at the top of the nav options, which currently list save and job poster only```

Suggested Additions:
- Back Button

joshdrink:
Chatted with @laurenoncoffee & @gobyrne - just a simple Save & Back button for managers when viewing an application.

gggrant:
Resolved by #828

PR #828, opened 2019-03-21
Adds Save & Back Button to HM Application View

joshdrink:
We added a back button the application view as a hiring manager to help facilitate navigation and forced this button to save data if there is unsaved data in the form directly below it.

#806


yonikid15:
I receiving a 500 error when hitting on the back button and being automatically redirected back to the applicants page

PR #832, opened 2019-03-22
Improve Laravel boot time

tristan-orourke:
I removed a bit of old unused code that was still part of the boot process, and refactored routes so they can be cached with `php artisan route:trans:cache`.

The performance gains on my local machine are very minor, but its worth trying on the live server anyway.


app/Http/Controllers/ApplicantProfileController.php
```diff
@@ -25,7 +25,7 @@ class ApplicantProfileController extends Controller
      * @param  \App\Models\Applicant $applicant
      * @return \Illuminate\Http\Response
      */
-    public function show(Request $request, Applicant $applicant)
+    public function show(Request $request, Applicant $applicant): \Illuminate\Http\Response
```
Khristophor: Be careful typehinting responses on controller actions, especially when returning a view. The Laravel view helper can return one of two different objects, and neither are of type `\Illuminate\Http\Response`.
If you put a mixed/optional typehint in the docblock, phpcs won't warn you about typehinting the function itself.
The danger comes from PHP just exploding if the typed return doesn't match what is actually returned at runtime.

app/Http/Controllers/ReferencesController.php
```diff
@@ -241,28 +241,27 @@ public function update(Request $request, ?Reference $reference = null)
...
+     * @param  \App\Models\Reference    $reference
      * @return \Illuminate\Http\Response
      */
     public function destroy(Request $request, Reference $reference)
     {
         $this->authorize('delete', $reference);
 
         //TODO: when projects exist independently on profile, delete seperatley
-        foreach($reference->projects as $project) {
+        foreach ($reference->projects as $project) {
```
Khristophor: Any chance we could utilize mass deleting here, instead of firing off individual delete calls in a loop?

From the Laravel docs:

>You can also run a delete statement on a set of models. In this example, we will delete all flights that are marked as inactive. Like mass updates, mass deletes will not fire any model events for the models that are deleted
```php
$deletedRows = App\Flight::where('active', 0)->delete();
```

app/Providers/AuthServiceProvider.php
```diff
@@ -8,18 +8,6 @@
 use GuzzleHttp\Client;
 use Lcobucci\JWT\Parsing\Decoder;
```
Khristophor: Are we using anything from this JWT package? We should also look to remove anything from composer.json we can.

routes/api.php
```diff
@@ -13,6 +13,6 @@
 |
 */
 
-Route::middleware('auth:api')->get('/user', function (Request $request) {
-    return $request->user();
-});
+// Route::middleware('auth:api')->get('/user', function (Request $request) {
```
Khristophor: Can we delete anything we're not using instead of commenting it out?

Khristophor:
Nice work, I love seeing PR's that are net negative 🥇

app/Providers/AuthServiceProvider.php
tristan-orourke: Good catch. I removed 3 more: `guzzle`, `jumbojett`, and `jwt`.

routes/api.php
tristan-orourke: Even the Laravel examples?

app/Http/Controllers/ApplicantProfileController.php
tristan-orourke: You're right that Response is the wrong type. I think its safe to use `\Illuminate\View\View` instead, as the view helper only returns the factory when you pass 0 arguments.

app/Http/Controllers/ReferencesController.php
tristan-orourke: Yep, it can be done here with `$reference->projects()->delete();`

routes/api.php
tristan-orourke: I believe its useful to have one example handy for each kind of route. Unless you feel very strongly, I've leaving these comments in for now.

tristan-orourke:
Ready for re-review

PR #842, opened 2019-03-26
Fix JobPoster index view in admin portal

Khristophor:


app/Models/JobPoster.php
```diff
@@ -233,6 +228,24 @@ public function submitted_applications() // phpcs:ignore
...
+     * withCount property on the model itself.
+     * See https://github.com/laravel/framework/issues/23957 for more info.
+     *
+     * @param mixed $value Value used to retrieve the model instance.
+     *
+     * @return \Illuminate\Database\Eloquent\Model|null
+     */
+    public function resolveRouteBinding($value) // phpcs:ignore
+    {
+        return $this->withCount('submitted_applications')->where('id', $value)->first() ?? abort(404);
```
tristan-orourke: Its worth double checking that $value is always an id and never the object itself. The cases that make me wonder are 
https://laravel.com/docs/5.8/redirects#redirecting-named-routes and 
https://laravel.com/docs/5.8/redirects#redirecting-controller-actions.

tristan-orourke: I played around with both of these, the route binding still works fine.

Issue #838, opened 2019-03-25
Bug - Job Poster Close Discrepancy During DST


joshdrink:
# Description
As it stands, job posters close at 23:59PM PST on the close date.

Daylight savings time complicates this, because job posters are actually closing at 01:59AM PDT during daylight savings time. Discussed with @laurenoncoffee and for now this is okay, but it will have to be fixed eventually so that job posters are always closing at midnight.

gggrant:
Resolved by #859

PR #859, opened 2019-04-02
Fix Job Poster open/close times not accounting for daylight savings

Khristophor:
Relates to #838 

Added new configuration setting for the `America/Vancouver` timezone. Automatically adjusts the UTC offset depending on the calendar date (ex. Mar 10 - Nov 3, 2019 will be UTC -7, otherwise UTC -8). All Job Posters should now open at 00:00:00 PST/PDT and close at 23:59:59 PST/PDT (this includes seeded posters).

Updated the model factory and Job Poster controller, as well as cleaned up some of the other factories.


gobyrne:
I like it. In most cases the "Open Time" won't matter, but you never know!

PR #878, opened 2019-04-03
Add default questions to admin-created job posters

### LINKED ISSUE #869: Bug - Deployment: Pre-populated "Fit" Questions Missing
# Description
It seems the two pre-populated questions we had on the job-poster form are no longer showing up (Gray and Lauren had two specific questions in this section that they always ask, and they are no longer there.

# Steps to Reproduce
1. Log in as an admin
2. Create a new job poster
3. Scroll down to "long form questions"
4. See no pre-populated questions

tristan-orourke:
Resolved #869

### DIFF
--- a/app/Http/Controllers/JobController.php
+++ b/app/Http/Controllers/JobController.php
@@ -232,6 +232,7 @@ class JobController extends Controller
         $jobPoster->manager_id = $manager->id;
         $managerEn = $manager->translate('en');
         $managerFr = $manager->translate('fr');
+
         $jobPoster->fill([
             'department_id' => $manager->department_id,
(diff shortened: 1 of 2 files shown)


app/Http/Controllers/JobController.php
```diff
@@ -245,6 +246,13 @@ public function createAsManager(Manager $manager)
             ]
         ]);
         $jobPoster->save();
+
+        $defaultQuestions = $this->populateDefaultQuestions();
+        debugbar()->debug($defaultQuestions);
```
Khristophor: Left a debugbar call here.

PR #881, opened 2019-04-04
Add skill_level langfiles back in

Khristophor:
Missing langfile(s) causing issues in the assessment plan. Looks like they were removed for failing a lang unit test comparison, as we never had French translations.


Khristophor:
@gobyrne we need French translations for this.

gobyrne:
Added french text, but didn't take a look at the page afterwards.

PR #903, opened 2019-04-12
Task/caps lock warning

### LINKED ISSUE #886: Sitewide - Caps Lock warning when entering password
# Description
Add a Caps Lock warning for when users accidentally have it on, when typing in passwords 
Example: https://www.w3schools.com/howto/howto_js_detect_capslock.asp

yonikid15:
Resolves #886 

### Notes:
- Made a caps lock warning during password input


resources/views/auth/login.html.twig
```diff
@@ -95,6 +95,7 @@
                                 required
                                 type="password"
                                 autocomplete="off" />
+                            <span class="password-caps-warning hidden">Warning: Caps lock engaged</span>
```
gggrant: This entry should be in a lang file with a French equivalent.

resources/views/auth/register.html.twig
```diff
@@ -151,11 +150,9 @@
                                         type="password"
                                         autocomplete="off" />
                                 </div>
-
                             </div>
-
                         </div>
-
+                        <span class="password-caps-warning hidden">Warning: Caps lock engaged</span>
```
gggrant: My password manager adds a div underneath the password field and obscures the text. Would above be better for most cases? Not a big deal.

yonikid15:
@gggrant are you sure it was working for me. It should appear when the password is in focus, the user is typing AND the caps lock is on. When the password input loses focus it will be set to hidden.

resources/views/auth/login.html.twig
yonikid15: Damn missed that sorry

yonikid15:
This is getting added to clone-framework

Issue #905, opened 2019-04-15
Bug - Elevating applicant to manager doesn't create a manager object


Khristophor:
# Description
When changing a user's role from applicant to manager in the admin portal, the newly elevated user does not appear in the admin portal list of manager accounts. We depend on a middleware to create the required manager object in the database when an account changes roles, but this only happens when that user logs in to the application. We need to add a mechanism to create this object during the role change process. Possibly a good candidate for [Laravel's event system](https://laravel.com/docs/5.8/events).

- [ ] This bug requires a unit test.

# Steps to Reproduce
1. Log in to the admin portal
2. Change a user's role from "applicant" to "manager"
3. Navigate to the Manager list in the admin portal
4. The changed user does not appear

PR #908, opened 2019-04-16
Manager not created after role change

Khristophor:
Relates to #905 

Gray noticed an issue yesterday where elevating an account from applicant to manager would not add them to the `Managers` section of the admin portal. When I looked into it I found that the middleware we assumed would handle this action only runs for the logged in user, and never fires unless an admin logs out and logs in as the new manager user.

In addition, elevating the account status ends up with a manager account that's missing required information for creating a job poster as that user account, so I added a custom exception and some more explicit messaging around the problem as an interim solution.


.vscode/settings.json
```diff
@@ -16,6 +16,9 @@
     { "language": "typescriptreact", "autoFix": true }
   ],
   "editor.formatOnSave": true,
+  "[php]": {
```
tristan-orourke: Can we keep this a user-defined setting?

app/Http/Controllers/JobController.php
```diff
@@ -234,18 +236,30 @@ public function createAsManager(Manager $manager)
...
+                    'division' => $managerEn->division,
+                ],
+                'fr' => [
+                    'branch' => $managerFr->branch,
+                    'division' => $managerFr->division,
+                ]
+            ]);
+            $jobPoster->save();
+        } catch (\Exception $e) {
+            throw new AdminException(
```
tristan-orourke: It should be possible for these to be left null. Initializing them to the manager's values is just a convenient default.

.vscode/settings.json
Khristophor: We could, but I don't really want anyone auto-formatting our PHP at this stage. It can introduce runtime errors when trying to properly guess typehints as we've seen.

app/Http/Controllers/JobController.php
Khristophor: Updated by moving more logic into the twig partial (which I'm not thrilled about, but it seems to be working okay).

tristan-orourke: I didn't realize this would require so much more work in the twig template, I'm sorry 😬

Khristophor: Haha no worries, just highlights more refactoring opportunity.

PR #913, opened 2019-04-18
Download all applications as PDFs

### LINKED ISSUE #860: Admin - (Concierge) Scripting Downloading All Applications for a Job
# Description
@gobyrne mentioned having one of us look into ways we might be able to script the process of downloading (printing as PDF) all the applications associated to a particular job poster so that it's super easy to send them as a package for HR.

The script might do things like:
- Take a job poster ID as a parameter
- Navigate to the correct page in the browser
- Open all applications for that job in new tabs
- CTRL P and Print as PDF for all tabs
- Save to a unique folder

gggrant:
Resolves #860.

### Notes:
- Requires creating a bookmarklet in your browser with a name like `Save PDF` and the following URL:
```javascript:function clickAllApplicants(){let t=document.querySelectorAll("div.box.lg-2of11.applicant-links > a:nth-child(1)");!function(t){(t=Array.prototype.slice.call(t)).map(function(t){return t.getAttribute("href")})}(t);for(var e=0;e<t.length;e++){var l=window.open(t[e],"_blank"),n=l.document.createElement("script");n.setAttribute("src",t[e].origin+"/js/getpdf.js"),l.document.body.appendChild(n)}}clickAllApplicants();```
- Run the bookmarklet on Review Applicants page

### CODE REVIEW

**Comment by tristan-orourke** [comment]
I moved the getpdf.js script into the resources folder, and loaded it into the public/js folder with webpack like everything else. The shortcut url will now need to be:
`javascript:function clickAllApplicants(){let t=document.querySelectorAll("div.box.lg-2of11.applicant-links > a:nth-child(1)");!function(t){(t=Array.prototype.slice.call(t)).map(function(t){return t.getAttribute("href")})}(t);for(var e=0;e<t.length;e++){var l=window.open(t[e],"_blank"),n=l.document.createElement("script");n.setAttribute("src",t[e].origin+"/getpdf.js"),l.document.body.appendChild(n)}}clickAllApplicants();`

File: resources/views/common/head.html.twig [hunk 1]
```diff
@@ -107,6 +107,11 @@
         crossorigin="anonymous">
     </script>
 
+{# jsPDF / html2canvas #}
+
+    <script src="https://html2canvas.hertzen.com/dist/html2canvas.js" integrity="sha384-vhfRL/EJVW2tOaB2/hQqMb8N29l78w84mJMs8WjKQ5ciRS9XsNeELFxLhVKoNKlQ" crossorigin="anonymous"></script>
```
tristan-orourke: The integrity check is failing for html2canvas, its blocked from downloading.

gggrant: New SRI hash added.

PR #960, opened 2019-05-13
Changed some of the client-side types, as part of normalization

tristan-orourke:
Removed nested Skill from the Criteria type, and some redundant english strings from some types containing both localizations..


tristan-orourke:
I waited too long to merge this. Its going to be a pain to reconcile with task/screening-plan-performance, so I'll do that after task/screening-plan-performance is merged in.

Issue #953, opened 2019-05-10
Manager - Server API routes for Job Poster


gggrant:
# Description
With validation, permissions.

PR #970, opened 2019-05-17
Feature/job api route

### LINKED ISSUE #952: Manager - Update Job Poster model
# Description
Backend model, DB migration, seeders, ensure existing pages/code/tests not broken

tristan-orourke:
Adds GET and UPDATE api routes for the job poster, with validation, policies, and unit tests in place. Adds several new fields to the JobPoster mode.

Resolves #952 and #953 .

### Notes:
- Depends on `feature/screening-plan-redux`. That branch must be merged into dev before this one.
- You must run `php artisan migrate` after merging this branch.


app/Http/Controllers/Api/JobApiController.php
```diff
@@ -0,0 +1,98 @@
...
+     *
+     * @param JobPoster $job
+     * @return mixed[]
+     */
+    private function jobToArray(JobPoster $job)
+    {
+        $criteria = Criteria::where('job_poster_id', $job->id)->get();
+        $criteriaTranslated = [];
+        foreach ($criteria as $criterion) {
+            // TODO: getTranslationsArray probably makes DB calls every loop. Find a way to profile & optimize.
```
Khristophor: Fun fact, it doesn't! I took a peek at the DB queries and this seems to be fine.

app/Http/Requests/UpdateJobPoster.php
```diff
@@ -0,0 +1,63 @@
...
+     *
+     * @return array
+     */
+    public function rules()
+    {
+        $dateFormat = Config::get('app.api_datetime_format');
+        $dateFormatRule = "date_format:$dateFormat";
+        return [
+            'term_qty' => 'required|numeric',
+            'open_date_time' =>["required", $dateFormatRule],
```
Khristophor: This is pretty nitpicky but can we settle on a quotes/double quotes standard for PHP? It's common to use singlequotes everywhere unless you're interpolating variables.

app/Http/Requests/UpdateJobPoster.php
```diff
@@ -0,0 +1,63 @@
...
+class UpdateJobPoster extends FormRequest
+{
+    /**
+     * Determine if the user is authorized to make this request.
+     *
+     * @return bool
+     */
+    public function authorize()
+    {
+        return true;
```
Khristophor: Is this superseded by the authorizeResource call in the controller?

app/Http/Controllers/Api/JobApiController.php
tristan-orourke: Yay!

app/Http/Requests/UpdateJobPoster.php
tristan-orourke: You're absolutely right.

app/Http/Requests/UpdateJobPoster.php
tristan-orourke: I believe it is, but I'll confirm.

tristan-orourke: We now have lots of options of where to run authorization: in the route file, in the controller, or in this Form Request. I liked having it all in the route file, so that all auth settings could be viewed in one place, but its the least flexible option. I've already had to break consistency and move some auth checks into controllers in some specific cases. Not sure what we should use going forward.

tristan-orourke: Just confirmed, there's a test for this. `JobApiControllerTest->testUpdateAsWrongManager()`

app/Http/Controllers/Api/JobApiController.php
```diff
@@ -0,0 +1,98 @@
...
+        return $jobArray;
+    }
+    /**
+     * Display a listing of the resource.
+     *
+     * @return \Illuminate\Http\Response
+     */
+    public function index()
+    {
+        //TODO: complete
```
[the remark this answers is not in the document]
tristan-orourke: Not yet, anyway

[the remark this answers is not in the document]
tristan-orourke: Looks like store needs to be added. Can I do that in a separate pull request?

PR #971, opened 2019-05-17
Progress Tracker Component

yonikid15:
Resolves #958 

### Notes:
- Merged with feature/screening-plan-redux, for jest setup
- Enzyme configuration added
- Components created and snapshot tested

PR #993, opened 2019-05-29
Job builder html templates

tristan-orourke:
These shouldn't be visible on the site anywhere yet, but they can be used as a starting point for development of the final Job Poster Builder pages.


tristan-orourke:
I want to merge this in, to make the changes in other pull requests, like #971, clearer

Issue #990, opened 2019-05-29
Bug - Manager - Screening Plan - Enter answer without criteria type


xander-miller:
## Description
A manager can start writing a model answer without selecting a criteria type, this creates error. Recommended fix: add logic to have answer field disabled until a criteria type is selected. 

- [x] This bug requires unit tests.

## Steps to Reproduce
Steps to reproduce the behavior. Example:
1. Add an Assessment type that requires questions
2. Add a question for that assessment type.
3. Add an answer but do not select a skill
4. Start typing an answer

## Screen Shot
![image](https://user-images.githubusercontent.com/663395/58582639-20101500-8220-11e9-81d9-b7f1c524963a.png)

Issue #985, opened 2019-05-27
Bug - Rating Guide Questions can be saved to server multiple times


tristan-orourke:
# Description
When a new Rating Guide Question is just created, its possible to cause multiple store requests to be launched. This causes one new question to suddenly multiply into several (on server and on client).

- [ ] This bug requires a unit test.

# Steps to Reproduce
Steps to reproduce the behavior. Example:
1. Go to Assessment Plan page.
2. Ensure there are some non-Narrative Review assessments.
3. Scroll down to Rating Guide Builder
4. As quick as possible, complete the following actions:
   a. add a question
   b. type a character into the question input
   c. click away from the question input, then back in
   d. type another character
   e. click away form the input again
5. If you do this fast enough, and then wait a few seconds, you should see several versions of the question appear.

Issue #983, opened 2019-05-27
Bug - You can add too many answers to a Rating Guide Question, if they have no skill selected


tristan-orourke:
# Description
You are only supposed to be allowed to add one Rating Guide Answer per related skill. Currently you can add as many as you want, if you leave skill unselected. Once you've selected a skill on enough of them, the extra answers vanish. (They will re-appear when the selected answers drop below the threshold again.) 

You should just be capped at the number of possible skills.

- [ ] This bug requires a unit test.

# Steps to Reproduce
Steps to reproduce the behavior. Example:
1. Go to Assessment Plan page.
2. Ensure there are some non-Narrative Review assessments.
3. Scroll down to Rating Guide Builder
4. Add a question.
5. Click the plus to add many answers. (More than the number of missing skills.)
6. Select a skill on answers. At some point, all possible answers will be selected, and extra answers will vanish.
7. Delete one answer with a selected skill, and the extras will reappear.
See error

# Images
Place image attachments here.

# Additional Context

PR #998, opened 2019-05-31
Performance Improvments and bug fixes for Screening Plan

### LINKED ISSUE #982: Bug - New Rating Guide Answers can cause errors on save
# Description
When you create a new Rating Guide Answer, if you add the answer text before selecting a skill, it will attempt to save to the server when focus is lost from the text input field. But, with skill_id null, it will cause an error on the server, and you'll see an error message:
```
Something went wrong!
Request failed with status code 400
```

- [ ] This bug requires a unit test.

# Steps to Reproduce
1. Go to an Assessment Plan page.
2. Ensure there are some assessment types besides  Narrative Review.
3. Scroll down to a Rating Guide Question (add a question if none exist).
4. Click the plus to add an answer (delete existing answers if the plus doesn't appear).
5. Edit the answer, without selecting a skill. Click away so the form input loses focus.
6. You should see an error message.

# Additional Context
Its caused because the server validates answers such that both answer text, and criterion id, are required.

tristan-orourke:
Significantly boost performance of screening plan by adding memoization to redux selectors and by moving reliance on redux data as low as possible into the component tree, in order to reduce the number of renders that happen. Previously, typing anything into a single input would cause the whole page to re-render.
Resolves #982, #983, #985, #990


resources/assets/js/components/AssessmentPlan/RatingGuideAnswer.tsx
```diff
@@ -102,7 +136,9 @@ const RatingGuideAnswer: React.FunctionComponent<
           onChange={(event): void =>
```
Khristophor: This is probably safe to leave for a future refactor, but we should stick to using named functions instead of anonymous ones (especially inside components that are repeated multiple times).

resources/assets/js/store/Assessment/assessmentSelector.ts
```diff
@@ -1,151 +1,212 @@
-import isEqual from "lodash/isEqual";
+import { isEqual } from "lodash";
```
Khristophor: Make sure to import the functions directly from lodash:
```js
import isEqual from "lodash/isEqual";
```
that case in particular saves about 65kb.

resources/assets/js/store/Job/jobSelectorComplex.ts
```diff
@@ -0,0 +1,209 @@
+import createCachedSelector from "re-reselect";
+import { uniq, difference, intersection } from "lodash";
```
Khristophor: Same thing here, you can import them all directly, although they have to be on individual lines.

resources/assets/js/components/AssessmentPlan/RatingGuideAnswer.tsx
tristan-orourke: For performance purposes?

app/Services/Validation/Rules/CriterionRule.php
```diff
@@ -0,0 +1,22 @@
...
+use App\Models\Criteria;
+
+/**
+ * Validates that value is a valid Criteria id
+ */
+class CriterionRule implements Rule
+{
+    public function passes($attribute, $value)
+    {
+        return Criteria::where('id', $value)->exists();
```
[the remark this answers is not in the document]
tristan-orourke: Oh yeah, this looks like it would be easy, actually. Don't know how it will compare performance wise, but it has to be minor.

resources/assets/js/components/AssessmentPlan/RatingGuideAnswer.tsx
tristan-orourke: I'll keep that in mind in the future

app/Http/Requests/StoreRatingGuideAnswer.php
```diff
@@ -0,0 +1,44 @@
...
+     * Determine if the user is authorized to make this request.
+     *
+     * @return bool
+     */
+    public function authorize()
+    {
+        // Ensure the user can make answers, and is the owner of the question it answers.
+        if ($this->user()->can('create', RatingGuideAnswer::class)) {
+            $questionId = $this->input("rating_guide_question_id");
+            if ($questionId) {
```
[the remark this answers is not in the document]
tristan-orourke: #0e7cbf05268667a52c3da631641108504b02f6fe

resources/assets/js/components/AssessmentPlan/RatingGuideAnswer.tsx
```diff
@@ -191,13 +257,29 @@ const mapDispatchToProps = (dispatch: DispatchType, ownProps): any => ({
...
 });
 // @ts-ignore
 const RatingGuideAnswerContainer: React.FunctionComponent<
   RatingGuideAnswerContainerProps
 > = connect(
   mapStateToProps,
   mapDispatchToProps,
-)(injectIntl(RatingGuideAnswer));
+  mergeProps,
+)(withPropsChecker(injectIntl(RatingGuideAnswer)));
```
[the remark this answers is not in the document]
tristan-orourke: 9452cc29999bf96805cf88b883d11299bc5e46e6

resources/assets/js/store/Assessment/assessmentSelector.ts
tristan-orourke: 08ab54cb35be20b3de917d97c1bf859c40799db4

resources/assets/js/store/Job/jobSelectorComplex.ts
tristan-orourke: 08ab54cb35be20b3de917d97c1bf859c40799db4

app/Services/Validation/Rules/CriterionRule.php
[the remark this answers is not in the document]
tristan-orourke: 24de2022ddb3ae50216e9927faedc5593fa471a8

PR #1002, opened 2019-06-03
JPB - Breadcrumbs Component

yonikid15:
### Notes:
- Snapshot tested
- Storybook documented


resources/assets/js/components/Breadcrumbs/BreadcrumbsItem.tsx
```diff
@@ -0,0 +1,22 @@
+import * as React from "react";
+
+interface BreadcrumbItemProps {
+  href: string;
+  title: string;
+}
+
+const BreadcrumbItem: React.FunctionComponent<BreadcrumbItemProps> = ({
```
tristan-orourke: This component is so small and specific I would have but it in Breadcrumbs.tsx.

resources/assets/js/components/Breadcrumbs/BreadcrumbsItem.tsx
```diff
@@ -0,0 +1,22 @@
...
+  href,
+  title,
+}): React.ReactElement => {
+  return (
+    <li>
+      <a href={href} title={title}>
```
tristan-orourke: Looks like, for accessibility, the current page needs to have a `aria-current="page"` property.

yonikid15:
@tristan-orourke do we still need this?

PR #1006, opened 2019-06-06
Update PHP formatting in controllers

Khristophor:
### Notes:
- Reviewed all controllers for explicit return typehints that can cause problems, and removed them
- Updated `ruleset.xml` with the following changes
  - Made sniffs relating to typehints warnings instead of errors
  - Excluded the `Controllers` directory from the typehint warnings
  - Added sniff for quote style and inline comment formatting
  - Changed all sniffs relating to "style", like commenting/docblock/etc warnings instead of errors
- Ran through a general formatting cleanup in all controllers

It should be safe to run PHP formatting on save now, but please pay attention to warnings in the IDE (green squigglies).

Sorry for the noise, but should be a good place to move forward from.

### DIFF
--- a/app/Http/Controllers/Admin/JobPosterCrudController.php
+++ b/app/Http/Controllers/Admin/JobPosterCrudController.php
@@ -14,8 +14,8 @@ class JobPosterCrudController extends CrudController
      */
     public function setup() : void
     {
-        $this->crud->setModel("App\Models\JobPoster");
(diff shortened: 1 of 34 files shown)


tristan-orourke:
Here's the guide to setting up php code sniffing: https://github.com/GCTC-NTGC/TalentCloud/wiki/%5BBED%5D-Onboarding-(Best-Practices)

Issue #968, opened 2019-05-17
Manager - Job Poster Builder Step 1 (Component only)


Khristophor:
# Description
Create a React FunctionalComponent for Job Poster Builder Step 1. This should have the complete, interactive ui, but not include a connection to redux. Use Storybook to mock the data and handlers that are passed in.

See #862 for more details on the ui.

  Storybook story.
  Jest snapshot testing.

Issue #862, opened 2019-04-02
Job Poster Builder: Step 02 - Job Information


joshdrink:
# Description
In this step, the hiring manager will provide basic job information like Job Title, Term, Classification, Level, Security Level, Language Profile, Location and are Remote Work, Telework and Flexible Hours allowed. After this step is complete we also show an encouraging render of some of the data.

As we add inputs to this UI, we will be removing the corresponding ones from the old UI that admins are still using (/en/manager/jobs/##/edit).

- [x] This task requires a unit test.

**Acceptance Criteria**
***Page & Nav***
- [ ] Depending on the button pressed on Step 0 (#942), we should either be on the /en version of the page or on the /fr version.
- [ ] If we are on the /en version of the page, all the "translatable" fields (like job title) should only be showing the English Input (and French only for /fr). This means you will need to toggle between the language of the page if you ever want to complete both languages.
- [ ] This page is the second on the progress tracker (top right) and should follow the same logic as described in #942

***Form***
- [ ] Job Title is the input for the EN or FR job title based on the language of the page you're on. This input should be removed from the old form (as all inputs we're moving to this form should be).
- [ ] "How long is the term (in months)" corresponds to Position Duration in the Timetable section of the old form.

Classification and Level are NEW. Previously this was a single text input with some regex restrictions. Now they have separate dropdowns.
- [ ] "What is the classification" should ideally get saved to the old text field we had for Classification. That will help us maintain backwards compatibility with older job posters.
- [ ] Classification should be removed from the old form.
- [ ] "What is the level" should be saved to a new integer field
- [ ] The Level pull-down is simply a list from 1 - 9
- [ ] The job poster needs to get updated to display this information correctly. It should display "Classification"-"Level". For example CS-04 or ENsur-05.

- [ ] "What is the Security Level?" replaces "Security Clearance" from the old form
- [ ] "What is the language profile" replaces "Language Requirement" from the old form.
- [ ] The text for "language profile" we want to display for managers is different then the text to display to candidates. Please see content doc for how they should appear here.
- [ ] Only NOC code should remain under "classifications" in the old form.

***Remote Work***
- [ ] As in the mock-up there is some added text below the input label.
- [ ] Remote work now has several options. On the old form it was just a checkbox. This new input should replace the old one.
[...]

PR #1008, opened 2019-06-06
Job Poster Builder step 01 components

Khristophor:
Affects #862, #968.

### Notes:
- Introduces Formik + Yup for easier form state management and validation
- Refactors the Modal component using React portals and context API for more flexibility with the markup and use cases

Still needs localization and tests.
Form element components should be migrated out of `JobDetails.tsx` into their own files for reuse, probably under a `Form` directory or similar.


yonikid15:
I have made a Form folder in the feature/job-builder/step-0 branch. Moved the Input, Select, etc. into it, while trying to make it a usable as a normal form element or a formik element.

resources/assets/js/components/JobDetails/JobDetails.tsx
```diff
@@ -0,0 +1,734 @@
...
+                      data-c-colour="white"
+                      data-c-font-size="h4"
+                      id="job-details-preview-title"
+                    >
+                      You&apos;re off to a great start!
+                    </h5>
+                  </div>
+                </Modal.Header>
+                <Modal.Body>
+                  <div
```
tristan-orourke: Extracting the Modal.Body content to a <JobPreviewCard> component might make it clearer what's going on, (and reduce this file size)

tristan-orourke: And allow it to be reused as the _actual_ job card later

resources/assets/js/components/Modal.tsx
```diff
@@ -1,189 +1,168 @@
-import React, { Component, createRef } from "react";
```
tristan-orourke: Meta comment about branch/PR hygene - maybe the changes to the Modal should (ideally) be its own PR

resources/assets/js/components/Modal.tsx
```diff
@@ -1,189 +1,168 @@
...
+    if (modalRef && modalRef.current) {
+      const focusableModalElements = modalRef.current.querySelectorAll(
+        'a[href], button, textarea, input[type="text"], input[type="email"], input[type="radio"], select',
+      );
+      const firstElement = focusableModalElements[0] as HTMLElement;
+      const lastElement = focusableModalElements[
+        focusableModalElements.length - 1
+      ] as HTMLElement;
+
+      if (!e.shiftKey && document.activeElement !== firstElement) {
```
tristan-orourke: If I understand correctly, this will always jump between the first and last element. That doesn't work if you add more focusable elements in the modal body.

resources/assets/js/components/Modal.tsx
```diff
@@ -1,189 +1,168 @@
...
+    }
+  }, [children]);
 
-  public handleClose = (event: React.MouseEvent<HTMLButtonElement>): void => {
-    event.preventDefault();
-    this.setState({ visible: false });
-    this.updateBody();
-  };
+  // Adds various key commands to the modal
+  useEffect((): (() => void) => {
```
tristan-orourke: Not criticism necessarily, but why does this need to rerun every time children change?

resources/assets/js/components/JobDetails/JobDetails.tsx
```diff
@@ -0,0 +1,734 @@
...
+        "flexHoursOccasionally",
+        "flexHoursSometimes",
+        "flexHoursNever",
+      ],
+      "Please select from the available options",
+    )
+    .required("This field is required."),
+});
+
+const TextInput = ({
```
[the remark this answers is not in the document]
Khristophor: Formik makes use of some extra props, like 
```js
  field: { name, value, onChange, onBlur },
  form: { errors },
```
we can look at making them optional I suppose, I thought it would be easier to have some separate form components

resources/assets/js/components/JobDetails/JobDetails.tsx
```diff
@@ -0,0 +1,734 @@
...
+    .min(2, "Too Short!")
+    .max(50, "Too Long!")
+    .required("This field is required."),
+  termLength: Yup.number()
+    .min(1, "Too Short!")
+    .max(36, "Too Long!")
+    .required("This field is required."),
+  classification: Yup.mixed()
+    .oneOf(
+      ["AS", "BI", "CO", "CR", "CS", "EC", "EX", "FO", "IS", "PC", "PE", "PM"],
```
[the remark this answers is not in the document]
Khristophor: I figured we'd be grabbing them from the backend at some point, mostly using this to mock out the frontend at this point.

resources/assets/js/components/Modal.tsx
Khristophor: It probably should be, although it was driven by the functionality in this story so I thought it was alright in this case.

resources/assets/js/components/Modal.tsx
Khristophor: On the first run, there aren't any children mounted, so no key commands are wired up correctly. I added the `children` to ensure the event handlers are bound correctly when the content of the modal has settled.

Khristophor: If new elements are added/removed from the modal that require focus, the event listeners should run again.

resources/assets/js/components/Modal.tsx
Khristophor: This just handles the cases where we need to "trap" focus in the modal. Outside of the first or last element in the map, it just tabs through normally.

resources/assets/js/components/JobDetails/JobDetails.tsx
Khristophor: Cool, I can do that.

resources/assets/js/components/JobDetails/JobDetails.tsx
```diff
@@ -0,0 +1,734 @@
...
+                    Next
+                  </button>
+                </div>
+              </Form>
+              <Modal
+                id="job-details-preview"
+                parentElement="#job-details"
+                visible={isModalVisible}
+                onModalConfirm={() => {
+                  console.log("Confirmed");
```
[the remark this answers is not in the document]
Khristophor: What would you recommend for the interim between components and full Redux functionality?

resources/assets/js/components/JobDetails/JobDetails.tsx
```diff
@@ -0,0 +1,734 @@
...
+                <Field
+                  name="classification"
+                  id="builder02Classification"
+                  label="What is the classification?"
+                  grid="tl(1of2)"
+                  component={SelectInput}
+                  required
+                  nullSelection="Select a classification..."
+                  options={[
+                    { value: "AS", label: "AS - Administrative Services" },
```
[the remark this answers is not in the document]
Khristophor: Same here, assumed API call for relevant options.

resources/assets/js/components/JobDetails/JobDetails.tsx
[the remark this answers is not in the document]
tristan-orourke: Hmm. Yoni did add the field and form props to the Input.tsx component on his branch. But the result is that most props are optional, and a bit confusing. I support having seperate Formik components that wrap the originals.

resources/assets/js/components/JobDetails/JobDetails.tsx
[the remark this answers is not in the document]
tristan-orourke: That makes sense. I'm okay with leaving this for this PR, and changing it when we hook it up to the backend.

resources/assets/js/components/JobDetails/JobDetails.tsx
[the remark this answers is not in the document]
tristan-orourke: I think JobDetails should accept handler functions as props, to handle things like submit and confirm. In the main app, we'll pass in redux action dispatchers, while the storybook could pass in console.log or anything else.

resources/assets/js/components/Modal.tsx
tristan-orourke: Yeah, I do this all the time, and I don't expect you to separate it out now. But with this bout of code reviews I've been wondering how we can improve our PR/branch strategies. And I think small feature branches that don't depend on anything except dev, that can be accepted quickly and easily, is the way to go.

resources/assets/js/components/JobDetails/JobDetails.tsx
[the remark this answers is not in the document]
yonikid15: https://github.com/GCTC-NTGC/TalentCloud/blob/feature/job-builder/step-0/resources/assets/js/components/Forms/Input.tsx 

This is what is looks like currently.

resources/assets/js/components/Modal.tsx
```diff
@@ -1,189 +1,168 @@
...
-        </div>
-      </>
-    );
-  }
+          {children}
+        </modalContext.Provider>
+      </div>
+    </div>,
+    parentElement && parentElement.length > 0
+      ? document.querySelector(parentElement)
```
tristan-orourke: It makes me nervous that we pass in a querySelector string into the modal, and hope it works. I'd rather pass in the dom node directly.

resources/assets/js/components/Modal.tsx
Khristophor: Nope, I'm wrong. Working on refactoring into something that works properly.

PR #1009, opened 2019-06-06
Get localized strings from models through a utility function

tristan-orourke:
It defaults to a fallback argument if the locale or field is missing.


resources/assets/js/components/AssessmentPlan/AssessmentPlan.tsx
```diff
@@ -17,7 +21,8 @@ const AssessmentPlan: React.FunctionComponent<
 > = ({ job, notifications, intl }): React.ReactElement => {
   const jobTitle = (
     <span data-c-colour="c5" data-c-font-size="h3">
-      {job && ` ${job[intl.locale].title}`}
+      {job &&
+        ` ${getTranslatedField(job, intl.locale, "title", "UNKNOWN JOB")}`}
```
Khristophor: I think I mentioned to you last week, but this is returning `UNKNOWN JOB` even when the API call resolves and the Job returned has a title available.

gggrant:
Stale. Keeping the branch, this may be implemented later.

PR #1016, opened 2019-06-07
Remove work environment demo code

gggrant:
Removed code from the team culture twig template. Part of the job builder merged early.


tristan-orourke:
We have to be careful with this. If it was merged early, and is undone in this commit, other PRs with this code won't add it back in.

PR #1020, opened 2019-06-10
Add email notifications to a queue, which will retry up to 5 times and log failures.

tristan-orourke:
The command `php artisan queue:work` starts the queue. Without running this command, emails will never be sent!
Alternatively, run the laravel task scheduler with a cron job. This will start the queue worker, and ensure it is restarted if it ever stops. https://laravel.com/docs/5.8/scheduling#introduction


database/migrations/2019_06_03_214715_create_jobs_table.php
```diff
@@ -0,0 +1,36 @@
+<?php
+
+use Illuminate\Support\Facades\Schema;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Database\Migrations\Migration;
+
+class CreateJobsTable extends Migration
```
Khristophor: This should probably be called `CreateQueuedJobsTable` for the sake of avoiding confusion.

app/Console/Kernel.php
```diff
@@ -39,4 +44,30 @@ protected function commands()
...
+    protected function osProcessIsRunning($needle)
+    {
+        // get process status. the "-ww"-option is important to get the full output!
+        exec('ps aux -ww', $process_status);
+
+
+        // search $needle in process status
+        $result = array_filter($process_status,
+            function($var) use ($needle)
+            {   return strpos($var, $needle); });
```
Khristophor: This block needs some formatting love:
```php
        $result = array_filter(
            $process_status,
            function ($var) use ($needle) {
                return strpos($var, $needle);
            }
        );
```

database/migrations/2019_06_03_214715_create_jobs_table.php
```diff
@@ -0,0 +1,36 @@
...
+    }
+
+    /**
+     * Reverse the migrations.
+     *
+     * @return void
+     */
+    public function down()
+    {
+        Schema::dropIfExists('jobs');
```
Khristophor: Probably don't want to drop the `jobs` table here 😆

Khristophor:
@tristan-orourke @gggrant What's the plan for running the email queue? Part of the deploy script or a cron setup?

PR #1040, opened 2019-06-18
W3C validation passes on all pages

gggrant:
### Notes:
- https://validator.w3.org/
- Used this extension for testing: https://www.validity.org.uk/


resources/views/common/modals/create_reference.html.twig
```diff
@@ -193,4 +194,4 @@
 
     </div>
 
-</div>
+<!--</div>-->
```
Khristophor: Curious why this closing tag is commented?

gggrant: Removed it, was a stray end tag on reference pages.

Khristophor: Are we sure this particular close tag is the culprit? This is a pretty small file with paired open/close tags... I have a suspicion the larger more complicated template that includes this one is the problem.

gggrant: I've restored it. Can't reproduce the stray tag error now.

PR #1057, opened 2019-06-27
Create/edit/delete departments via admin panel

### LINKED ISSUE #1031: Sitewide - Bring Department Impact Statements into the Database
## Description
Department Impact Statements are part of the information about a department and should not be hard coded into the front end of the app. Move Department Impact Statements into the database. 

- [x] This task requires a unit test.
- [x] This task is complex and needs acceptance criteria.

### LINKED ISSUE #1032: Admin - Make Department Partners Editable by Admins
## Description
The List of department will change every year. Admins should be able add/remove/edit the departments without the aid of developers. Make the department list editable through the admin portal.

- [x] This task requires a unit test.
- [x] This task is complex and needs acceptance criteria.

gggrant:
Resolves #1031 and #1032.


database/migrations/2018_12_21_155421_populate_departments.php
```diff
@@ -14,39 +14,39 @@ class PopulateDepartments extends Migration
     public function up()
     {
         DB::table('departments')->insert([
-            ['id' => 1, 'name' => 'treasury_board'],
```
tristan-orourke: This is to avoid screwing up the auto-increment index, right? But we have to be careful, this edited version will never run on prod, since this migration has already ran.

database/migrations/2019_07_02_191541_add_translated_name_and_impact_to_departments.php
```diff
@@ -0,0 +1,52 @@
...
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('departments', function (Blueprint $table) {
+            $table->json('name')->nullable();
+            $table->json('impact')->nullable();
+        });
+
+        $departments = Department::all();
```
tristan-orourke: Its a bad idea to reference the Department model in a migration, because the model could change later, breaking this migration. Its better to use raw db queries. For an example, take a look at database\migrations\2019_06_11_220005_convert_existing_work_env_and_culture_to_job_poster.php

resources/lang/en/common/lookup/departments.php
```diff
@@ -10,14 +10,54 @@
...
-    "natural_resources" => "Natural Resources Canada",
-    "transport" => "Transport Canada",
-    "enviroment_and_climate_change" => "Environment and Climate Change Canada",
-    "employment_and_social_development" => "Employment and Social Development Canada",
-    "global_affairs" => "Global Affairs Canada",
-    "fisheries_and_oceans" => "Fisheries and Oceans Canada",
-    "innovation_science" => "Innovation, Science and Economic Development Canada",
-    "public_service_and_procurement" => "Public Services and Procurement Canada",
-    "border_services_agency" => "Canada Border Services Agency",
+    'departments' => [
```
tristan-orourke: If the only place these values are used is the AddTranslatedNameAndImpactToDepartments migration, its probably better to define these values in that migration file.

resources/lang/fr/common/lookup/departments.php
```diff
@@ -10,14 +10,55 @@
...
-    "natural_resources" => "Ressources naturelles Canada",
-    "transport" => "Transports Canada",
-    "enviroment_and_climate_change" => "Environnement et Changement climatique Canada",
-    "employment_and_social_development" => "Emploi et Développement social Canada",
-    "global_affairs" => "Affaires mondiales Canada",
-    "fisheries_and_oceans" => "Pêches et Océans Canada",
-    "innovation_science" => "Innovation, Sciences et Développement économique Canada",
-    "public_service_and_procurement" => "Services publics et Approvisionnement Canada",
-    "border_services_agency" => "Agence des services frontaliers du Canada",
+    'departments' => [
```
tristan-orourke: See above. This file should be deleted.

app/Http/Controllers/DepartmentController.php
```diff
@@ -0,0 +1,34 @@
...
+use App\Models\Lookup\Department;
+
+class DepartmentController extends Controller
+{
+    /**
+     * Return all departments as an array
+     *
+     * @return mixed
+     */
+    public function index()
```
tristan-orourke: If we're adding this controller, we might as well create an api route while we're at it.

database/migrations/2018_12_21_155421_populate_departments.php
tristan-orourke: We'll need to reset the auto-increment either manually, or with a new migration.

Khristophor: Why do we need to reset the auto increment? Are we referencing departments by ID somewhere?

database/migrations/2019_07_02_191541_add_translated_name_and_impact_to_departments.php
gggrant: Refactored.

resources/lang/en/common/lookup/departments.php
gggrant: Removed.

resources/lang/fr/common/lookup/departments.php
gggrant: Removed.

database/migrations/2018_12_21_155421_populate_departments.php
gggrant: I think this is resolved.

tristan-orourke: We have departments with ids up to 12, but the auto-increment index is still at 1. So here's the issue: when someone tries to add a new department from the admin portal, it will fail with a "Duplicate key" error.

database/migrations/2019_07_04_090953_update_department_translations.php
```diff
@@ -0,0 +1,130 @@
...
+    /**
+     * Reverse the migrations.
+     *
+     * @return void
+     */
+    public function down()
+    {
+        Schema::create('department_translations', function (Blueprint $table) {
+            $table->dropColumn('impact');
+            $table->getChangedColumns('value');
```
tristan-orourke: What's this line for?

PR #1107, opened 2019-07-16
Adds NRC to departments table

gggrant:
### Notes:
- Includes all current partner departments now.


xander-miller:
Pretty straight forward migration to add a department.

Issue #959, opened 2019-05-13
Job Poster Builder: Step 05 - Tasks


nicolsonaitken:
The fifth step in creating a job poster is Tasks; Hiring Managers will be asked to create a minimum of four tasks and a maximum of eight key tasks needed to accomplish their position. Once these are completed, we show the hiring manager a preview on the following page. The preview should match the updated Tasks (Your Work) section of the job poster.

- [x] This task requires a unit test.
- [x] This task is complex and needs acceptance criteria.

**Acceptance Criteria**

***User Field(s)***
- [ ] On page load there is one Task input area consisting of an up icon, a down icon, a text area and a delete button.
- [x] On page load the up, down and delete icons are disabled.
- [x] The Add Task button will add the task as well as adding a new User Input Field underneath the previous task.
- [x] Each task remains editable within the list.
- [x] When there are two or more tasks, the up icon will move the Task up one position within the list of tasks. 
- [x] When there are two or more tasks, the down icon will move the Task up one position within the list of tasks. 
- [x] The delete icon will delete the task from the list.
- [x] If a sixth task is added, an error message will appear notifying the user that the maximum threshold has been reached, however a new User Input Field will also appear.
- [x] Each task after the sixth will be highlighted as an error (red border on the input field).
 
***Counter***
- [x] On page the counter says 'You’ve haven’t added any Key Tasks yet.'
- [x] With the addition of the first task the counter will change to 'You’ve added 1 Key Task'
- [x] As each new task is added the number will increase and Key Task will become Key Tasks

***Preview***
- [x] The "Preview" button takes you to the preview page
- [x] The preview should show this section as it will appear on the job poster.
- [x] If there are seven or more task then the preview will take the user to an error message letting them know they need to trim their tasks to six or under. 

Link to wireframes:
[Tasks](https://xd.adobe.com/view/ad49a964-1211-4533-6cf8-115df49b5414-db78/)

Link to content:

[Tasks](https://docs.google.com/document/d/1ifJapgBuBHv7-M-B3ULwNiEYSchQ1_GePp4OHplpzDE/edit?usp=sharing)

PR #1119, opened 2019-07-18
Add Job Poster Builder Tasks page (component only)

Khristophor:
Affects #959


resources/assets/js/components/JobTasks/JobTasks.tsx
```diff
@@ -1,222 +1,387 @@
...
+                                          description="Error message displayed when too many tasks are on screen."
+                                          defaultMessage="You have exceeded the maximum number of key tasks allowed, but that’s okay. You can continue to add key tasks as you brainstorm here, but you will be asked to trim your list to 6 key tasks or fewer to proceed."
+                                        />
+                                      </p>
+                                    </div>
+                                  </div>
+                                </div>
+                              )}
+                              <div
+                                key={`listItem${index}`}
```
tristan-orourke: Its dangerous to use index as the key, _particularly_ when the order will change. Is it possible to give new tasks a temporary unique id somehow? 
https://reactjs.org/docs/lists-and-keys.html#keys

resources/assets/js/components/JobTasks/JobTasks.tsx
```diff
@@ -1,222 +1,387 @@
...
+                name="tasks"
+                render={(arrayHelpers): React.ReactElement => (
+                  <>
+                    <div data-c-grid="gutter">
+                      {values.tasks &&
+                        values.tasks.length > 0 &&
+                        values.tasks.map(
+                          (task, index): React.ReactElement => (
+                            <>
+                              {validCount! === index && (
```
tristan-orourke: Why the non-null assertions on validCount, here and later? Don't we need to account for it possibly being undefined?

resources/assets/js/components/JobTasks/JobTasks.tsx
```diff
@@ -1,222 +1,387 @@
...
+                                  >
+                                    <i
+                                      className="fas fa-trash"
+                                      data-c-colour="stop"
+                                    />
+                                  </button>
+                                </div>
+                                <Field
+                                  id={`task${task.id}`}
+                                  name={`tasks.${index}.description`}
```
tristan-orourke: I would have expected using index like this not to work, but it tests fine in storybook. I guess arrayHelpers makes it work?

resources/assets/js/components/JobTasks/JobTasks.tsx
Khristophor: After reviewing the functionality for the page, I'm making it required. I'll take out the non-nulls.

resources/assets/js/components/JobTasks/JobTasks.tsx
Khristophor: It's a combo of Formik allowing it and having a helper function to grab the nested object (if it exists). I had to update the TextAreaInput component to make use of it, but it seems to work fine for array nested inputs as well as individual ones.

https://jaredpalmer.com/formik/docs/api/fieldarray#fieldarray-array-of-objects

resources/assets/js/components/JobTasks/JobTasks.tsx
Khristophor: I'll see what I can do. We'll have to figure out some sort of temp ID standard, unless we're just worried about having it as the key in isolated cases like this.

Khristophor: As we discussed I added the [nanoid](https://github.com/ai/nanoid) library for generating temp IDs. We'll keep it to 10 characters for now as that seems to be more than fine.

I'm still seeing console warnings for missing keys inside of a loop, so I'm not sure if we'll still see unnecessary rerenders inside the tree. I think it has to do with how I'm using fragments inside the FieldArray but no combination of adding keys to any element inside of those loops seem to resolve it.

PR #1143, opened 2019-07-22
Feature/job builder/update job detail modal content

xander-miller:
Updates Job Preview Modal Content and adds localization to Job Preview Modal Content


resources/assets/js/components/JobPreview.tsx
```diff
@@ -11,36 +17,112 @@ export interface JobPreviewProps {
...
   securityLevel,
-  startDate,
   classification,
+  telework,
   level,
-}: JobPreviewProps): React.ReactElement => {
+  intl,
+}: JobPreviewProps & InjectedIntlProps): React.ReactElement => {
   return (
     <div
```
Khristophor: Looks like the majority of the markup has changed for this component, see
https://github.com/GCTC-NTGC/TalentCloud/blob/feature/job-builder/step-03-ui/resources/views/manager/builder-02.html.twig#L455-L521

resources/assets/js/components/JobPreview.tsx
```diff
@@ -68,67 +150,183 @@ const JobPreview: React.FunctionComponent<JobPreviewProps> = ({
...
+        />
       </h4>
       <div data-c-grid="gutter">
         <div data-c-grid-item="tp(1of2)">
           <p data-c-colour="c1" data-c-margin="bottom(quarter)">
-            Average Annual Salary
+            <FormattedMessage
+              id="jobPreview.lengthOfTheTerm"
+              defaultMessage="Length of the Term"
+              description="Datum Label"
```
Khristophor: I know it's kind of annoying to have to fill out all of these descriptions, but I believe the spirit of the property is to add context to the line being translated. When a translator receives a file with hundreds of items needing translation, it's not very clear what the context is surrounding the lines.

Having `Datum Label` in this case for most of the translations in this file isn't super helpful to that end.

resources/assets/js/components/JobPreview.tsx
```diff
@@ -68,67 +150,183 @@ const JobPreview: React.FunctionComponent<JobPreviewProps> = ({
...
+        </div>
+        <div data-c-grid-item="tp(1of2)">
+          <p data-c-colour="c1" data-c-margin="bottom(quarter)">
+            <FormattedMessage
+              id="jobPreview.travel"
+              defaultMessage="Travel"
+              description="Datum Label"
+            />
+          </p>
+          <p>Not yet implemented.</p>
```
Khristophor: This should be localized and/or conditionally shown based on the prop.

resources/assets/js/components/JobPreview.tsx
```diff
@@ -68,67 +150,183 @@ const JobPreview: React.FunctionComponent<JobPreviewProps> = ({
...
+        </div>
+        <div data-c-grid-item="tp(1of2)">
+          <p data-c-colour="c1" data-c-margin="bottom(quarter)">
+            <FormattedMessage
+              id="jobPreview.overtime"
+              defaultMessage="Overtime"
+              description="Datum Label"
+            />
           </p>
+          <p>Not yet implemented.</p>
```
Khristophor: This should be localized and/or conditionally shown based on the prop.

resources/assets/js/components/JobDetails/JobDetails.tsx
```diff
@@ -367,6 +368,7 @@ const jobToValues = (job: Job | null, locale: string): JobFormValues =>
         securityLevel: job.security_clearance_id || "",
         language: job.language_requirement_id || "",
         city: job[locale].city || "",
+        education: job[locale].education || "",
```
Khristophor: Education isn't actually part of the form, so it doesn't really need to be set in JobToValues or UpdateJobWithValues

resources/assets/js/components/JobDetails/JobDetails.tsx
```diff
@@ -950,7 +955,10 @@ const JobDetails: React.FunctionComponent<
                         province={intl.formatMessage(
                           provinceName(Number(values.province)),
                         )}
+                        education={values.education}
```
Khristophor: Education should come from the Job object here

resources/assets/js/components/JobDetails/JobDetails.tsx
```diff
@@ -950,7 +955,10 @@ const JobDetails: React.FunctionComponent<
...
                         termLength={Number(values.termLength)}
+                        telework={String(values.telework)}
```
Khristophor: I think it makes sense to pass the already localized versions of telework and flexhours here, rather than duplicating all the translations within the JobPreview component.

It also tightly couples the preview to the JobDetails form.

resources/assets/js/components/JobPreview.tsx
```diff
@@ -11,36 +17,112 @@ export interface JobPreviewProps {
...
   classification: string;
   /** Level for the classification, i.e. 03 */
   level: string;
+  /** How often are Flex Hours allowed */
+  flexHours: string;
+  // Description of Required Education
+  education: string;
+  // Telework is allowed
+  telework: string;
+  // Travel opportunities;
```
Khristophor: Is this for another prop?

resources/assets/js/translations/locales/fr.json
```diff
@@ -109,6 +108,24 @@
   "formValidation.tooLong": "Too long!",
   "formValidation.tooShort": "Too short!",
   "job.daysSinceClosed": "{dayCount, plural, =0 {Aucun jour} one {# jour} other {# jours}} depuis la fermeture",
+  "jobBuilder.culturePace.01.description": "Our deadlines are tight, we balance several tasks at the same time, and our priorities are always changing. Our work should come with running shoes!",
```
Khristophor: Most of these look like translations for other files, they should be committed as part of another PR (or multiple PR's)

resources/assets/js/components/JobPreview.tsx
xander-miller: Agreed I would like something that provides context to the header. how is? "Job Preview Card Info Label"

resources/assets/js/components/JobPreview.tsx
xander-miller: Yeah, I was trying to split some of the differences between the old design and the new design. It was a stupid move in hind sight. Fixed. 👍

resources/assets/js/translations/locales/fr.json
xander-miller: That's a result of translations being added with out the `npm run translation` script being run. I will make a separate PR that is just updating the translations. That should fix this.

xander-miller: Done. 👍

resources/assets/js/components/JobDetails/JobDetails.tsx
xander-miller: What if the form is being used to update an existing job poster? I thought in that case the job information should already be there and passed through.

resources/assets/js/components/JobPreview.tsx
xander-miller: Agreed, conditionally shown on the prop. +1

resources/assets/js/components/JobPreview.tsx
xander-miller: Agreed, conditionally shown on the prop. 👍

resources/assets/js/components/JobDetails/JobDetails.tsx
xander-miller: Yeah, now that I take a step back and think about it. It doesn't make sense. 👍

resources/assets/js/components/JobDetails/JobDetails.tsx
xander-miller: Makes sense 👀 Done 👍

resources/assets/js/components/JobDetails/JobDetails.tsx
xander-miller: Agreed. I'm using a different approach now. 👍

resources/assets/js/components/JobPreview.tsx
xander-miller: Yes, and that prop is back. 👍

PR #1145, opened 2019-07-22
Feature/user softdeletes

### LINKED ISSUE #1144: Feature - Applicant/Manager(?) - Account deletion
# Description

Until we add a self-service account deletion feature, admins should be allowed to soft-delete users.

gggrant:
Resolves #1144.

### Notes:
- Migration to add soft delete column to users table.
- Added SoftDeletes to User model
- Allow delete on User admin panel


gggrant:
Ditching this interim solution, sorting out the account deletion requests manually.

An account deletion feature is forthcoming.

PR #1156, opened 2019-07-24
Classifications admin panel

### LINKED ISSUE #1158: Admin - Classifications admin panel
# Description
After updating the skill admin panel, the ability to add new classifications will be useful.

gggrant:
Resolved #1158.

### Notes:
- Allows admins to create/edit occupational groups (classification codes). 
- Anything missing?


Khristophor:
@gggrant Looks good to me, although we also have a couple Classification fields within the Job Poster table (`classification`, which I think is only there for backwards compatibility, and `classification_code`, which is the two letter designation). Are we planning on making use of the classifications table for Job Posters as well, or just Skills for now?

tristan-orourke:
> @gggrant Looks good to me, although we also have a couple Classification fields within the Job Poster table (`classification`, which I think is only there for backwards compatibility, and `classification_code`, which is the two letter designation). Are we planning on making use of the classifications table for Job Posters as well, or just Skills for now?

Referencing the classification table on job_posters makes sense, to enforce validity. I set this up in b7700777fdef3f66c8eb7fb05d84e1b0cab8215b, without changing the API.

PR #1190, opened 2019-08-01
Feature/word counter

### LINKED ISSUE #917: Applicant - Word Counter (for skill declarations)
# Description
Follow up on PR #915 and #874 

- The word counter still needs a few more things before it can be added in:

   - [ ] Info message correlating to amount of words
   - [ ] what the max/goldilock/min amount of words should be

# Wireframe
- https://xd.adobe.com/view/e79e0e9e-d1ae-4b0c-749d-9436424f1ae7-91af/

yonikid15:
Resolves #917 

### Notes:
- Word Counter for Skills page
- Run Storybook to view OR on the Applicant Side - Step 3 and 4 of application process


resources/assets/js/components.tsx
```diff
@@ -8,6 +8,6 @@
 // require('./components/ReferenceList');
 require("./components/ApplicationReview/ReviewApplicationsContainer");
 require("./components/ApplicationReview/ApplicationReviewContainer");
-require("./components/ApplicantSkills/WordCounter");
+require('./components/WordCounter/WordCounter');
```
tristan-orourke: First, I believe WordCounter should be replaced with WordCounterWrapper now.

But more importantly, we've decided recently to try to break our js into seperate bundles, to reduce the amount of unnecessary code downloaded by the browser for a given page. In this case, it means that _instead_ of adding WordCounter to `components.tsx`, you should export WordCounterWrapper in `webpack.mix.js`, and include this specific script in the twig template. Check `job-builder-intro.html.twig` for an example of how this is done.

resources/assets/js/components/ProgressRing.tsx
```diff
@@ -1,4 +1,4 @@
-import React from "react";
+import React from 'react';
```
tristan-orourke: It appears you have different autoformat settings that I do. We should all be using the same config, and it should be using double-quotes. Lets look into this.

resources/assets/js/components/WordCounter/WordCounter.test.ts
```diff
@@ -0,0 +1,61 @@
+import {
+  countNumberOfWords,
+  stringEndsInWhitespace,
+  truncateWords,
+} from "./WordCounter";
```
tristan-orourke: This file is now testing the WordCounter utility file, now the component itself. Imports need to change, and probably the describe wordings.

resources/assets/js/components/WordCounter/WordCounter.tsx
```diff
@@ -0,0 +1,106 @@
...
+*/
+
+interface Message {
+  max: number;
+  message: string;
+}
+
+interface WordCounterProps {
+  numOfWords: number;
+  maxWords: number;
```
tristan-orourke: I'd add some comments for the interface fields. For example, the difference between maxWords, numOfWords, and wordLimit is not obvious. Also add a comment about the message array needing to be ordered from lowest max to largest.

tristan-orourke: Although perhaps this component could itself reorder messages by max, just to be sure.

resources/assets/js/components/WordCounter/WordCounter.tsx
```diff
@@ -0,0 +1,106 @@
...
+  maxWords,
+  warnings,
+}): React.ReactElement => {
+  const message = (): string => {
+    let index = 0;
+    const warning = warnings.find(
+      ({ max }, i): boolean => {
+        index = i;
+        return (
+          warnings[i + 1] &&
```
tristan-orourke: If you assume warnings are ordered from smallest max to largest (which I think has to be true for this to work), do you need to compare to `warnings[i+1]` at all?

resources/assets/js/components/WordCounter/WordCounterWrapper.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+      if (target.value.trim()) {
+        setCurrentNumberOfWords(numOfWords);
+      } else {
+        setCurrentNumberOfWords(0);
+      }
+
+      if (numOfWords < wordLimit) {
+        // After maxLength is set to positive number, it cannot be set back to unlimited (-1)
+        element.maxLength = 1000000000;
```
tristan-orourke: I don't think this is required anymore.

resources/assets/js/components/WordCounter/helpers.ts
```diff
@@ -0,0 +1,27 @@
+export const countNumberOfWords = (innerText: string): number => {
```
tristan-orourke: I like separating these functions out into their own file 👍

resources/views/common/skill.html.twig
```diff
@@ -572,8 +572,8 @@
                                     {# Word Counter #}
                                     {# <div
                                         data-word-counter-id={{ "#{id_prefix}skillDescription"|json_encode() }}
-                                        data-max-words={{ "100"|json_encode()  }}
-                                        data-min-words={{ "75"|json_encode()  }}
+                                        data-max-words={{ "80"|json_encode()  }}
```
tristan-orourke: We'll want to set the max and min words in the same place we create the messages, (speaking of which, we'll need translations for those) which probably means in a very context-specific SkillDeclarationWordCounter component. This component will need to pass in all the constants to WordCounterWrapper and also mount the component with ReactDOM.

resources/assets/js/components/WordCounter/types.ts
```diff
@@ -0,0 +1,23 @@
+export interface WordCounterProps {
```
tristan-orourke: These don't need to be in their own file.

tristan-orourke: In fact, component props should _definitely_ be in the same file as their component.

resources/assets/js/stories/WordCounter/WordCounter.stories.tsx
```diff
@@ -0,0 +1,22 @@
...
+import WordCounterWrapper from "../../components/WordCounter/WordCounterWrapper";
+import SkillsWordCounter from "../../components/ApplicantSkills/SkillsWordCounter";
+
+const stories = storiesOf("Components|Word Counter", module)
+  .addDecorator(withInfo)
+  .addDecorator(withKnobs)
+  .addDecorator(withIntl);
+
+stories.add(
+  "Word Counter",
```
tristan-orourke: I added another story for WordCounterWrapper so you can adjust the word threshold with knobs.

resources/assets/js/translations/extractedMessages/resources/assets/js/components/ApplicantSkills/SkillsWordCounterMessages.json
```diff
@@ -0,0 +1,27 @@
+[
```
tristan-orourke: This file should be deleted, shouldn't it?

yonikid15: Ya, that's wierd.

resources/assets/js/components/ApplicantSkills/SkillsWordCounterMessages.ts
```diff
@@ -0,0 +1,53 @@
+import { defineMessages } from "react-intl";
+
+export const wordCounterMessages = defineMessages({
+  skillsMessage_1: {
+    id: "wordCounter.skills.skillsMessage_1",
```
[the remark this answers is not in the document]
Khristophor: I was also renaming these 😆

resources/assets/js/components/WordCounter/types.ts
yonikid15: Okay!

resources/assets/js/components/ApplicantSkills/SkillsWordCounterMessages.ts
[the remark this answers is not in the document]
yonikid15: We already renamed these and added it into the SkillsWordCounter.tsx component.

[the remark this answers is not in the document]
Khristophor: Too late :P

PR #1198, opened 2019-08-02
Update CI configuration to reduce runtime

Khristophor:
### Notes:
Trying to cut down the Travis/Scrutinizer run time, as we seem to be doubling up on a lot of checks across different services.

### DIFF
--- a/.scrutinizer.yml
+++ b/.scrutinizer.yml
@@ -42,7 +42,3 @@ filter:
 checks:
   php: true
 
-tools:
-  external_code_coverage:
-    timeout: 777
-    runs: 1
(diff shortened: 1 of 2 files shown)


Khristophor:
So, we can cut down Scrutinizer time pretty significantly (5 minutes vs 13 minutes on an older inspection) if we go with this config and don't wait for code coverage reports from Travis. If we _do_ want to incorporate code coverage, it'll negatively impact times across all the PR's. This leads to a massive queue backup, since Scrutinizer can only use one container at a time.

The might be some room for improvement in the Travis script but that's a can of worms I don't really want to get into right now.

Also, considering I don't think anyone really looks at the Scrutinizer results, it might be worthwhile to switch to something like [Larastan](https://github.com/nunomaduro/larastan) that does local static analysis for Laravel projects. I tried it out locally and it processed over 200 files in milliseconds, so we'd be able to add it to Travis without much impact on the build time.

Issue #1010, opened 2019-06-06
Job Poster Builder: Step 08 - Review


JasmitaChawla:
## Description
The final step for the job poster builder, this section shows managers how their job poster will look to applicants. 

- [ ] This task requires a unit test.
- [ ] This task is complex and needs acceptance criteria.

## Acceptance criteria

- [ ] There's an edit button under each section that once clicked, takes the user back to the step where the information for that section was entered.

- [ ] After users make a change and click preview, there's an additional third button on the modal "Go to Review" that allows them to skip straight to this review page (as opposed to clicking through all the steps).

- [ ] On this page, the entire work environment section is displayed. But for the applicants, when the job poster is live, there should be a link that opens the full section on a separate page:

How poster should look like to applicants:

<img width="375" alt="Screen Shot 2019-06-13 at 10 18 06 AM" src="https://user-images.githubusercontent.com/38079220/59440204-8f5f3a80-8dc4-11e9-93ca-0e50dab3103b.png">

Clicking on the "View the team’s work environment and culture" link should open a separate page with the full section:

<img width="334" alt="Screen Shot 2019-06-13 at 10 19 46 AM" src="https://user-images.githubusercontent.com/38079220/59440339-cd5c5e80-8dc4-11e9-975c-661398f56d11.png">

## Link to prototype
[Prototype
](https://xd.adobe.com/view/85e7b2e5-3ca1-4a14-4f62-82ddfba3c63c-f3d6/)

## Link to copy
[Copy](https://docs.google.com/document/d/1P28kVMYARFDeg3RB_rBPRq8g4oLULq206BcXVHvRne4/edit?usp=sharing)

PR #1232, opened 2019-08-08
Added a basic Middle Btn option to the Modal component

tristan-orourke:
### Notes:
- This is being added now to enable the Skip to Review buttons for #1010.


tristan-orourke:
I thought I could give each button `data-c-grid-item="base(1of3)"`, but now the main Confirm button is not enough to the right when the middle button is not present.

PR #1266, opened 2019-08-13
Bugfix - JPB Skills page 

yonikid15:
Resolves #issue during sprint review

### Notes:
- On the skills page there's a modal that has a single button. The **children** prop returns a object instead of an array when there is only 1 item. Therefore, I added a condition making sure **children** is an array before mapping and adding the **gridItem** prop.


resources/assets/js/components/Modal.tsx
```diff
@@ -161,22 +161,24 @@ Modal.Footer = function ModalFooter(props): React.ReactElement {
...
-              gridItem: children.length,
-            });
-            return btnWithGridItemProp;
-          },
-        )}
+        {Array.isArray(children) && children.length > 0
+          ? children.map(
+              (btn): React.ReactElement => {
+                const btnWithGridItemProp = React.cloneElement(btn, {
+                  gridItem: children.length,
```
Khristophor: Seeing a warning in the console about `gridItem`

```
Warning: React does not recognize the `gridItem` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `griditem` instead. If you accidentally passed it from a parent component, remove it from the DOM element.
```

yonikid15: Changed it to lowercase 👍

PR #1295, opened 2019-08-19
Client-Side routing between Job Builder pages

tristan-orourke:
Creates a JobBuilderRoot component that uses client-side routing to render all the JobBuilder steps. This allows quick navigation between them without reloading data from the server API each time.

### Notes:
-[ react-router](https://github.com/ReactTraining/react-router) is used elsewhere in the project. I used [ Universal Router](https://github.com/kriasoft/universal-router)+ [History ](https://github.com/ReactTraining/history) here instead because they allow for much simpler, easier to read routing, and combined still make up a smaller package. I wrapped use of them in a router.ts utility file roughly based on [hookrouter](https://github.com/Paratron/hookrouter). This should make it relatively easy to replace Universal Router with a different implementation, if we want.


resources/assets/js/components/JobBuilder/JobBuilderRoot.tsx
```diff
@@ -0,0 +1,69 @@
...
+      },
+      {
+        path: "/:id/builder/skills",
+        action: ({ params }) => <JobSkillsPage jobId={Number(params.id)} />,
+      },
+      {
+        path: "/:id/builder/review",
+        action: ({ params }) => <JobReviewPage jobId={Number(params.id)} />,
+      },
+    ],
```
Khristophor: The "submit for review" step is throwing a console error due to the route not being found.

resources/assets/js/components/JobReview/JobReviewPage.tsx
```diff
@@ -73,11 +74,11 @@ const JobBuilderReviewPage: React.FunctionComponent<
 
   const handleReturn = (): void => {
     // Go to Previous page
-    window.location.href = jobBuilderSkills(locale, jobId);
+    navigate(jobBuilderSkills(locale, jobId));
   };
   const handleContinue = (): void => {
     // Continue to next page
-    window.location.href = managerJobIndex(locale);
+    navigate(managerJobIndex(locale));
```
Khristophor: I guess this should be reverted to preserve the right functionality?

tristan-orourke: Yes, looks like I got over-enthusiastic with the navigate()'s

PR #1358, opened 2019-08-28
Avoid null exceptions when trying to publish poster.

### LINKED ISSUE #1355: Bug - Admin - Publishing a Job causes an error if open date time hasn't been set
# Description
Publishing a job from the admin portal causes an exception if the job does not have an open date set yet.
This action should be disabled, or at least fail gracefully with an informative message, if critical fields haven't been set.

# Steps to Reproduce
1. Create a job in the Job Poster Builder.
2. Complete, and submit for review.
3. Logout, and log in to admin portal.
4. Find the job you just submitted for review.
5. Click edit on the job row.
6. Click publish, and save.
7. See error.

tristan-orourke:
Resolves #1355.

### Notes:
- This is a very minimal PR. It doesn't stop an admin from publishing a Job in an incomplete state (missing open dates, close dates, etc). That can be added in a separate PR.

### DIFF
--- a/app/Http/Controllers/Admin/JobPosterCrudController.php
+++ b/app/Http/Controllers/Admin/JobPosterCrudController.php
@@ -80,6 +80,14 @@ class JobPosterCrudController extends CrudController
             ]
         ]);
         $this->crud->addField([
+            'name' => 'open_date_time',
+            'label' => 'Open Date',
+            'type' => 'datetime_picker',
+            'datetime_picker_options' => [
+                'format' => 'YYYY-MM-DD HH:mm:ss',
+            ],
+        ]);
+        $this->crud->addField([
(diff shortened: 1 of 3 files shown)


tristan-orourke:
@Khristophor I've simplified the failing test, in line with the simplified mutator. I can't find a good reason why published_at date was tied to open_date in the first place.

Khristophor:
> @Khristophor I've simplified the failing test, in line with the simplified mutator. I can't find a good reason why published_at date was tied to open_date in the first place.

Not sure either, honestly. One of those things that made sense at the time I guess. The rationale seemed to be "If it's past the open date, and it hasn't been published yet, publish it now with the current timestamp. If it's not past the open date, we want it to be published as soon as it's open (so the dates would be the same)."

PR #1362, opened 2019-08-29
Social Cards Prep / on Job Index & Posters

joshdrink:
Adds a social card template to the platform that can be added to any page that uses the base layout. These will be added to pages as needed - the homepages will get them when the new homepage designs go live.


gggrant:
Reopen if/when the project team wants to move forward on this.

PR #1363, opened 2019-08-29
Made JPB the default form for Managers. Old form trimmed down, moved to Admin portal.

### LINKED ISSUE #1340: Task - Manager - Job Poster Builder Link
## Description

On the manager Job poster index view add a create new job poster link and change edit job poster links to use the Job Poster Builder. 

## Images

image

## Acceptance criteria
- [ ] Add Create New Job Poster Link
- [ ] Change Edit Job Poster Link to use Job Poster Builder

tristan-orourke:
Resolves #1340.

### Notes:
- Managers have a Create Job button again!
- Manager's 'edit job' links bring them to the JPB
- Manager's "Submit for Review" also bring them to the JPB, but directly to the submit for review button. This is easier for now, because the validation code already exists on the review step, and not on the original button.
- Removed all fields from the old form that are also on the JPB
- Old form only accessible from the Admin portal now.
- Admins now get a link to the jpb and the old form, and have permissions to edit either.


routes/web.php
```diff
@@ -403,9 +385,15 @@ function () : void {
         'middleware' => ['auth', 'role:admin']
     ],
     function (): void {
-        Route::get('jobs/create/as-manager/{manager}', 'JobController@createAsManager')
-            ->middleware('can:create,App\Models\JobPoster')
-            ->name('admin.jobs.create.as_manager');
+        /* Edit Job */
```
Khristophor: Do these not need to be localized? I know right now it's probably only one or two people using the admin portal for edits, but we have the translations handy. It also seems to be affecting other routing with the menus and such (logout triggers a 405 from these pages currently).

tristan-orourke: You're right, it makes more sense to localize this page.
I'll also fix the menu, so it acts like its on the admin portal, not the applicant portal.

PR #1386, opened 2019-09-04
Bugfix - JPB - Remove back button validation

### LINKED ISSUE #1342: Bug - Manager - Job Poster steps back button validations
## Description
Each step of the Job Poster builder will not allow you to use the save and back button if there are required fields that have not been filled out yet. Create a yup validation without required field validation and apply it to the submit of each back button. 

## Acceptance criteria
- [ ] Job Info Back Button works without required fields
- [ ] Work Environment back button works without required fields
- [ ] Impact back button works without required fields
- [ ] Tasks back button works without required fields
- [ ] Skills back button works without required fields

yonikid15:
Resolves #1342 

### Notes:
- Removed removed validation on JPB forms when moving to previous step.
- However, all values should still be saved when moving to previous step.


resources/assets/js/components/JobBuilderWorkEnv/WorkEnvForm.tsx
```diff
@@ -882,7 +893,7 @@ const WorkEnvForm = ({
           />
         </h4>
         <p data-c-margin="bottom(normal)">
-          <FormattedMessage {...formMessages.ourWorkEnvDesc} />
+          {intl.formatMessage(formMessages.ourWorkEnvDesc)}
```
Khristophor: Just curious why the swap from `FormattedMessage` to `intl.formatMessage`

yonikid15: I was getting an eslint error using the spread syntax.

Khristophor: Ah okay, I thought I had added an ignore for that globally but it must be in my draft PR

yonikid15: Is Formatted Message preferable?

Khristophor: It doesn't really matter, I was just wondering since they both do the same thing

PR #1400, opened 2019-09-06
New colour scheme, applicant homepage, compliant headers/footers

joshdrink:
I do need someone to dev the "date modified" aspect of the footer before this gets merged. @tristan-orourke @Khristophor 

Otherwise, it would be good to get a visual once over to make sure I haven't busted anything up.


resources/views/applicant/home.html.twig
```diff
@@ -7,23 +7,229 @@
...
+    <div data-clone>
+        {% include "common/goc" %}
+        {% include "common/alert" %}
+        <div
+            data-c-background="white(100)">
+            {# Home Heading ================================================= #}
+            <header data-c-align="center" data-c-overlay="black(80)" data-c-padding="bottom(triple)" style="background-image: url('/images/bg_crowd.jpg');">
+                {% include "common/menu" %}
+                <div data-c-container="large" data-c-padding="top(double)">
+                    <h1 data-c-colour="white" data-c-font-size="h1" data-c-font-weight="700" data-c-margin="top(triple) bottom(normal)">{{ home.hero.title }}<img alt="Canadian Maple Leaf" class="hero-icon" src="/images/maple-leaf.svg"></h1>
```
Khristophor: alt tags should be localized as well

resources/views/applicant/application_index.html.twig
```diff
@@ -11,9 +11,8 @@
 
 {% block body %}
 
-	{% include "common/alert" %}
-
-    {% include "common/goc" %}
+	{% include "common/goc" %}
```
Khristophor: This is more of a refactoring note, but if we're making the same change a dozen times (in this case reordering two components), these views should all be extending a template that contains these headers.

PR #1434, opened 2019-09-12
Update mix config and React bundles

Khristophor:
### Notes:
- Replaced moment with dayjs
- Removed generic `components.tsx`
- Added new entrypoints for existing React pages (screening plan, applicant review)
- Removed duplicate jquery ui file
- Moved redux_index to an assessment_plan view
- Updated clone CDN link to be versionless
- Moved `mix.version();` into an environment check (production only)


tristan-orourke:
Some react components use Axios to make api requests. These components were previously included in bootstrap.js, and bootstrap.js could ensure that it configured axios properly before those scripts ran. Now that these components have their own entry points into the app, the config needed to happen differently: 646fd002557ed2e4bc78ca8ad0b0a0e4d02bfb6b

webpack.mix.js
```diff
@@ -56,3 +72,7 @@ mix.webpackConfig({
     },
   },
 });
+
+if (mix.inProduction()) {
```
tristan-orourke: I don't see the advantage of only versioning in production. Old cached js files can cause plenty of confusion in dev environments.

PR #1443, opened 2019-09-13
Collect more info on Manager Registration

tristan-orourke:
Resolves #1274.

### Notes:
- When registering as a manager, we also collect department and Gov email. (Alternatively, user states "I'm not from the government".)
- When an applicant uses the Manager portal for the first time, they must complete the same information before continuing.

Issue #1275, opened 2019-08-14
Manager Demo - Email Management & Government Email Field on Profile


joshdrink:
# Description
On their profiles, managers need to be able to edit their email/user name, as well as their new "government email" field from the registration (see issue #1274)

PR #1447, opened 2019-09-13
Manager Profile - Added password reset form and form validation

### LINKED ISSUE #1047: Manager Demo - Bug - Manager's don't have a form for updating their passwords
# Description
The form where applicants can update their password is not currently available to hiring managers. They need some way to do this (other than the "reset password" button).

- [ ] This bug requires a unit test.

yonikid15:
Resolves #1047 & #1275 

### Notes:
- Added new section to Manager Profile called "My Account"
    - Moved name field from "About Me" to "My Account" section.
    - Added new 'government email' field
- Added form validation.

NOTE: PR #1443 should get merged in first.


database/migrations/2018_07_12_145513_create_managers_table.php
```diff
@@ -15,6 +15,7 @@ public function up()
 		Schema::create('managers', function(Blueprint $table)
 		{
 			$table->increments('id');
+			$table->string('government_email')->nullable();
```
Khristophor: I don't think we can edit a previous migration with a new column, it won't get run in production

resources/views/manager/profile/about.html.twig
```diff
@@ -4,6 +4,7 @@
     Manager: Profile - About Me
 
 ============================================================================= #}
+{% set linkedInPattern = '^https:\/\/[a-z]{2,3}\.linkedin\.com\/.*$' %}
```
Khristophor: This should be set somewhere central, like a config file or inside a validator, and passed in via the controller

resources/views/manager/profile/layout.html.twig
```diff
@@ -4,6 +4,13 @@
     Manager - Profile Layout
 
 ============================================================================= #}
+{% set managerEN = manager.translate('en') %}
```
Khristophor: I'd like these to be set in the controller and passed in as well, but that's more of a personal preference thing

app/Http/Controllers/Api/ManagerApiController.php
```diff
@@ -68,14 +68,14 @@ public function showAuthenticated()
     /**
      * Update the specified resource in storage.
      *
-     * @param  \App\Http\Requests\UpdateManager $request Incoming Form Request.
-     * @param  \App\Models\Manager              $manager Incoming Manager.
+     * @param  \App\Http\Requests\UpdateManagerApi $request Incoming Form Request.
```
Khristophor: Curious why this was renamed?

yonikid15: Asked by Tristan, I believe to prevent confusion between the new UpdateManagerProfileRequest.php file.

app/Http/Controllers/ManagerProfileController.php
```diff
@@ -6,9 +6,14 @@
...
 use Illuminate\Http\Request;
 use App\Http\Controllers\Controller;
+use App\Http\Requests\UpdateManagerProfileRequest;
 use App\Models\Lookup\Frequency;
 use App\Models\Lookup\Department;
 use App\Models\Manager;
+use App\Services\Validation\Rules\LinkedInUrlRule;
+use App\Services\Validation\Rules\TwitterHandlerRule;
+use Illuminate\Support\Facades\Hash;
+use TijsVerkoyen\CssToInlineStyles\Css\Rule\Rule;
```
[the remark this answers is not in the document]
yonikid15: Passing the pattern to the input field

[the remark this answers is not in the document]
Khristophor: `TijsVerkoyen\CssToInlineStyles\Css\Rule\Rule`?

app/Http/Requests/UpdateManagerProfileRequest.php
```diff
@@ -0,0 +1,118 @@
...
+
+    /**
+     * Determine if the user is authorized to make this request.
+     *
+     * @return boolean
+     */
+    public function authorize()
+    {
+        $manager = $this->route('manager');
+        return $manager->id !== null;
```
tristan-orourke: 1. If the Request class found a manager object, it did so using an id. That means that if `$manager` is not null, it definitely has an id. You'd want to check `$manager !== null`, not `$manager->id !==null`, which will either return true or throw an exception.
2. I don't think this check is necessary at all. Just by typehinting the Manager class in the controller route, it will fail if its null.

app/Http/Requests/UpdateManagerProfileRequest.php
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    /**
+     * Override Handle a failed validation attempt.
+     *
+     * @return void
+     */
+    protected function failedValidation(\Illuminate\Contracts\Validation\Validator $validator)
+    {
+        $this->validator = $validator;
```
tristan-orourke: What's the purpose of overriding this?

app/Http/Requests/UpdateManagerProfileRequest.php
```diff
@@ -0,0 +1,118 @@
...
+        return [
+            'name' => 'required|string|max:191',
+            'email' => [
+                'required',
+                'string',
+                'max:191',
+                'email',
+                // Email may match existing email for this user,
+                // but must be unique if changed.
+                Rule::unique('users', 'email')->ignore($this->manager->user->id)
```
tristan-orourke: Are you sure `$this->manager` will work in this class?

app/Services/Validation/Rules/TwitterHandleRule.php
```diff
@@ -0,0 +1,43 @@
...
+     * This regex will allow only alphamumeric characters and the underscore.
+     * Keep this handy if we need to validate other usernames.
+     *
+     * @param  string  $attribute
+     * @param  mixed   $value
+     * @return boolean
+     */
+    public function passes($attribute, $value)
+    {
+        return preg_match('/' . self::PATTERN . '/', $value);
```
tristan-orourke: Since we know twitter handles have a max length, we should fold that into this rule as well.

app/Services/Validation/Rules/TwitterHandleRule.php
```diff
@@ -0,0 +1,43 @@
...
+    }
+
+    /**
+     * Get the validation error message.
+     *
+     * @return string
+     */
+    public function message()
+    {
+        return Lang::get('validation.regex');
```
tristan-orourke: You could create a custom message, saying "This is not a valid twitter handle".

resources/views/manager/profile/about.html.twig
```diff
@@ -138,10 +106,21 @@
                             {% for department in departments %}
                                 {% set deparmentName = department.name %}
                                 {% set chosenName = manager.department.name %}
+                                {% if old('department_id') == department.id %}
```
tristan-orourke: If `old('department_id')` is different from chosenName, you could end up with two options selected!

app/Http/Requests/UpdateManagerProfileRequest.php
yonikid15: Seems to be working.

app/Http/Requests/UpdateManagerProfileRequest.php
yonikid15: This is so I have access to the `$request->validator->fails()` method. With that I can scroll to errors section. 

```php
// redirect to error messages element if validation fails
        if (isset($request->validator) && $request->validator->fails()) {
            $hash = '#managerProfileFormErrors';
            return redirect(route('manager.profile.edit', $manager).$hash)
                        ->withErrors($request->validator)
                        ->withInput();
        }
```

app/Services/Validation/Rules/TwitterHandleRule.php
```diff
@@ -0,0 +1,43 @@
...
+use Illuminate\Support\Facades\Lang;
+
+class TwitterHandleRule implements Rule
+{
+    /**
+     * Validation for Twitter handlers.
+     *
+     * @var string
+     */
+    const PATTERN = '^[A-Za-z0-9_]{1,10}$';
```
tristan-orourke: Doesn't this limit it to 10 characters? It should be 15. While you're at it, mention the length limit in the docstring below.

app/Http/Requests/UpdateManagerApi.php
```diff
@@ -29,21 +31,20 @@ public function rules()
...
+                'string',
+                'max:191',
+                'email',
+            ],
             'twitter_username' => [
-                'nullable', //Some people may not have a handle.
-                'max:15', //Per Twitter's Terms/Service.
-                'regex:/^[A-Za-z0-9_]+$/',
+                'nullable', // Some people may not have a handle.
+                'max:15', // Per Twitter's Terms/Service.
```
tristan-orourke: Since the Twitter rule checks the length, remove the max here.

resources/views/manager/profile/about.html.twig
```diff
@@ -129,7 +97,7 @@
                             required
                             class="form__input form__input__select"
                             id="managerProfileDepartment"
-                            name="department">
+                            name="department_id">
 
                             <option value="" disabled {% if chosenName %}selected{% endif %}>
```
tristan-orourke: Looks like you're using chosenName before you've defined it.

resources/views/manager/profile/about.html.twig
```diff
@@ -138,12 +106,21 @@
                             {% for department in departments %}
                                 {% set deparmentName = department.name %}
                                 {% set chosenName = manager.department.name %}
-                                <option
-                                    value="{{ department.id }}"
-                                    {% if deparmentName == chosenName %}selected="selected"{% endif %}
-                                    >{{ department.name }}</option>
+                                    <option
```
tristan-orourke: This is super nitpicky, but while we're here, the the indentation in this block is a bit off :P

PR #1559, opened 2019-09-30
Fix bug where gov email wasn't appearing on First Manager Visit page.

### LINKED ISSUE #1558: Bug - Gov Email input not appearing on the First Manager Visit screen
# Description
When first visiting the Manager portal as an applicant user, you're redirected to a screen to enter your department, and (if you don't select "I'm not from the Government") your government email. The government email input isn't appearing, but you can't submit the form.

# Steps to Reproduce
1. Register on the applicant portal.
2. Once logged in, navigate to /manager.
3. Select a department (not "I'm not from the Government")
4. Attempt to continue by clicking submit. If you open the chrome inspector, you should see a console error.

# Images
Place image attachments here.

- [x] Regression test written.

tristan-orourke:
Resolves #1558.

### DIFF
--- a/resources/views/auth/manager_registration_fields.html.twig
+++ b/resources/views/auth/manager_registration_fields.html.twig
@@ -1,0 +1,45 @@
+{# =============================================================================
+
+    This partial template contains 2 fields, department and gov_email, which
+    are used when registering as a manager.
+
+    It requires the /js/managerRegistration.js script in order to function properly.
+
+============================================================================= #}
+
+<div
+    data-c-grid-item="base(1of1)"
+    data-c-input="select"
+>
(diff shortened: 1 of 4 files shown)


resources/views/auth/manager_registration_fields.html.twig
```diff
@@ -0,0 +1,45 @@
+{# =============================================================================
+
+    This partial template contains 2 fields, department and gov_email, which
+    are used when registering as a manager.
+
+    It requires the /js/managerRegistration.js script in order to function properly.
```
Khristophor: Shouldn't the script be included with this partial if this is the case?

PR #1598, opened 2019-10-02
Bugfix - JPB single-page app navigation

### LINKED ISSUE #1332: Bug - Client-side navigate() method stores incorrect page title in history.
# Description
When navigating between steps of the Job Builder, every page is added to browser history with the same Page Title as the original page you navigated to.

# Steps to Reproduce
1. Got to `/en/manager/jobs/builder`.
2. Complete the intro step, click Continue.
3. Complete the Details form, save and preview, click Continue.
4. Long press on the browser back button: both previous entries appear as `Job Builder: Intro`.

# Images
Place image attachments here.

yonikid15:
Resolves #1332 

### Notes:
- Set document title and h1 tag with router on react side.
- Moved translations to React-side.


resources/assets/js/helpers/router.ts
```diff
@@ -37,17 +37,21 @@ export const useUrlHash = (): void => {
 };
 
 export const useRouter = (
-  routes: Routes<any, React.ReactElement>,
+  routes: Routes<any, any>,
```
tristan-orourke: You can easily define the type as `{title: string; component: React.ReactElement}`

resources/assets/js/helpers/router.ts
```diff
@@ -37,17 +37,21 @@ export const useUrlHash = (): void => {
...
   const [component, setComponent] = useState<React.ReactElement | null>(null);
 
   // Render the result of routing
   useEffect((): void => {
-    router
-      .resolve(location.pathname)
-      .then((result): void => setComponent(result));
+    router.resolve(location.pathname).then((result): void => {
+      const { defaultMessage } = result.title.props;
+      const h1 = document.getElementById("h1");
```
tristan-orourke: It seems odd to me to use "h1" as an id. Why not just search for the first <h1> element?

Khristophor:
```js
document.querySelector('h1');
```

yonikid15: Lol. For some reason I thought you only had access to innerHTML with getElementById.

resources/assets/js/helpers/router.ts
```diff
@@ -37,17 +37,21 @@ export const useUrlHash = (): void => {
...
   const router = useMemo(() => new UniversalRouter(routes), [routes]);
   const [component, setComponent] = useState<React.ReactElement | null>(null);
 
   // Render the result of routing
   useEffect((): void => {
-    router
-      .resolve(location.pathname)
-      .then((result): void => setComponent(result));
+    router.resolve(location.pathname).then((result): void => {
+      const { defaultMessage } = result.title.props;
```
tristan-orourke: You're passing in an entire `FormattedMessage` React element in as `result.title`, but you're only using the props of that react element. You might as well just pass the props directly.

Actually, You're only grabbing defaultMessage from those props, which is always the English string (it doesn't get translated). I think you should try to translate the message before it gets here, so that title can be a simple string.

resources/assets/js/components/JobBuilder/JobBuilderRoot.tsx
```diff
@@ -13,47 +14,110 @@ import JobSkillsPage from "../JobBuilderSkills/JobBuilderSkillsPage";
...
-      },
-      {
-        path: "/:id/builder/tasks",
-        action: ({ params }) => <JobTasksPage jobId={Number(params.id)} />,
-      },
-      {
-        path: "/:id/builder/skills",
-        action: ({ params }) => <JobSkillsPage jobId={Number(params.id)} />,
+        action: () => ({
+          title: <FormattedMessage {...titles.introTitle} />,
```
tristan-orourke: `FormattedMessage` is a ReactElement that can only translate its contents if it's below [IntlProvider](https://github.com/formatjs/react-intl/blob/master/docs/Components.md#intlprovider) in the React element tree. It won't be much help for setting document.title. Instead, you should call intl.formatMessage to convert it directly to a string. You could change `routes` from a constant to a function that accepts the intl object, which JobBuilderRoot can pass in.

Khristophor:
> The react-intl v3 PR is about to be merged in. This will require a few changes after that.

I think we might be safe on this actually. The `FormattedMessage` component didn't change at all, and `defineMessages` is just an object.

resources/assets/js/components/JobBuilder/JobBuilderRoot.tsx
Khristophor: You can use the [new hook](https://github.com/formatjs/react-intl/blob/master/docs/API.md#useintl-hook) to grab the intl object now

app/Http/Controllers/JobBuilderController.php
```diff
@@ -11,83 +11,76 @@ class JobBuilderController extends Controller
      * Show the Job Builder Intro page
      * @return \Illuminate\Http\Response
      */
-    public function intro($jobId = null)
+    public function intro()
     {
         return view(
-            'manager/job-builder-root',
-            ['title' => Lang::get('manager/job_builder.intro_title'), 'jobId' => $jobId]
```
tristan-orourke: Good point, this isn't necessary any more.

resources/assets/js/helpers/router.ts
```diff
@@ -36,21 +38,33 @@ export const useUrlHash = (): void => {
   }, [location.hash, hashFound]);
 };
 
-export const useRouter = (
-  routes: Routes<any, React.ReactElement>,
-): React.ReactElement | null => {
+export const UseRouter: FunctionComponent<{
```
tristan-orourke: I don't really like the change of UseRouter to a component. Or rather, if it is going to be changed to a component, it should not be named like a hook and should be moved to to the components folder. I'm curious why you made the change in the first place.

resources/assets/js/helpers/router.ts
```diff
@@ -36,21 +38,33 @@ export const useUrlHash = (): void => {
...
+  routes: Routes<any, any>;
+  scrollToTop?: boolean;
+}> = ({ routes, scrollToTop }): React.ReactElement | null => {
+  const intl = useIntl();
   const location = useLocation();
   const router = useMemo(() => new UniversalRouter(routes), [routes]);
   const [component, setComponent] = useState<React.ReactElement | null>(null);
+  const tracker: HTMLElement | null = document.getElementById(
+    "job-builder-root",
+  );
```
tristan-orourke: This is specific to the Job Builder, and breaks this hook/component's potential reusability.

resources/assets/js/helpers/router.ts
yonikid15: I remember running into multiple problems with the routes array and adding the intl hook (which needs a FC, and has to have the IntlProvider component as an ancestor) to properly translate the titles. However, I refactored everything which should resolve everything without having to make UseRouter a FC.

app/Http/Controllers/JobBuilderController.php
```diff
@@ -11,83 +11,76 @@ class JobBuilderController extends Controller
      * Show the Job Builder Intro page
      * @return \Illuminate\Http\Response
      */
-    public function intro($jobId = null)
+    public function intro()
```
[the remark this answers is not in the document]
tristan-orourke: Yes, we can get rid of them. At this point, we can even get rid of the controller, and just return the view straight from routes. https://laravel.com/docs/6.x/routing#view-routes

resources/assets/js/helpers/router.ts
tristan-orourke: Nicely done.

PR #1646, opened 2019-10-07
Add logging and improve toast message for 419 errors.

### LINKED ISSUE #1619: Bug - Manager portal JPB page not loading in IE after editing a step
# Description

When using a demo account on the manager portal, on the job poster builder, on the review page, clicking on any of the edit links, making a change in that step, then clicking save and preview, the page doesn't load. This only happens in IE and not in Chrome. 

# Steps to Reproduce
1. Log on to the manager portal as a demo account user using Internet Explorer.
2. Draft a job poster.
3. On the Review step, click on any of the edit links.
4. Then click on "Save and Preview" on the bottom of the page. The page doesn't load. 

# Images

tristan-orourke:
Resolves #1619.

### Notes:
- This doesn't _really_ resolve #1619, but we suspect that bug was caused by a 419 Error that is difficult to reproduce. It should help us gather more information about this inconsistent bug. In the meantime, the ErrorToast will now instruct users to refresh the page, which should allow them to continue.


app/Exceptions/Handler.php
```diff
@@ -40,6 +58,18 @@ class Handler extends ExceptionHandler
...
+        if ($exception instanceof TokenMismatchException) {
+            $logData = [
+                'requestToken' => request()->header('x-csrf-token'),
+                'sessionToken' => session()->token(),
+                'session' => session()->all(),
+                'user' => request()->user(),
+                'requestUrl' => request()->url()
+            ];
+            $message = '419 CSRF Token Mismatch. ' . collect($logData)->toJson();
+            Log::debug($message);
```
Khristophor: Will `Log::debug()` run outside of a local environment?

app/Exceptions/Handler.php
```diff
@@ -55,6 +85,11 @@ public function render($request, Exception $exception)
         if ($exception instanceof AdminException) {
             return $exception->render($request);
         }
+        if ($exception instanceof TokenMismatchException) {
+            $newMessage = $exception->getMessage() . " Please refresh the page.";
```
Khristophor: We'll need to localize this

tristan-orourke: Good point.

app/Exceptions/Handler.php
Khristophor: Just took a gander at the source, doesn't look like it checks for any sort of levels for different environments, so you are indeed correct

PR #1654, opened 2019-10-08
Add support for two factor authentication

### LINKED ISSUE #1569: Spike - Choose how to implement 2 Factor Authentication
# Description
First we need to solidify our second factor. We've discussed making certain routes only available on secure networks, texting single use codes, or using an app for single use codes.

Second, we need to pick which technologies, services, libraries, etc we'll use to implement 2FA, if any.

Khristophor:
Resolves #1569 

### Notes:
- Uses [google2fa-laravel](https://github.com/antonioribeiro/google2fa-laravel) and [BaconQrCode](https://github.com/Bacon/BaconQrCode)
- Provides a middleware for wrapping routes
- Still a WIP but I have tested locally with the Google Authenticator app and can add a code that saves to the user account
- Need to implement a view for the OTP and test out things like recovery codes/deactivation
- Server requires [Imagick](https://www.php.net/manual/en/book.imagick.php) to be installed to generate the QR code (PECL package [here](https://pecl.php.net/package/imagick).)


Khristophor:
After chatting with @tristan-orourke, it makes sense to refactor the TwoFactorController to use the User object rather than the Applicant as it is now. Will need some route refactoring since I'm passing the applicant ID through the url to the controller.

app/Http/Controllers/Auth/LoginController.php
```diff
@@ -84,6 +84,9 @@ public function showLoginForm()
      */
     public function logout(Request $request)
     {
+        // Invalidates any existing 2FA verified devices.
+        $request->user()->cycleRememberDeviceToken();
```
Khristophor: I think I'd like to remove this before going live, although we might want to add a "forget remembered devices" button on the settings page that calls `cycleRememberDeviceToken()`.

Thoughts @tristan-orourke ?

tristan-orourke:
> I think I'd like to remove this before going live, although we might want to add a "forget remembered devices" button on the settings page that calls cycleRememberDeviceToken().

Yep, good idea. I made those changes, and added success messages (following Grant's pattern) for Deactivate 2FA and Forget Trusted Devices.

Also, it seemed like remember token was only being saved if the user object happened to be saved elsewhere in the same request, so I made sure RememberDeviceTrait always saves the token to DB. But I'm confused about why it was working at all, previously, so please sanity check this @Khristophor cd080293fd8c3d44a04173f905d599c88e05022a

Khristophor:
> > I think I'd like to remove this before going live, although we might want to add a "forget remembered devices" button on the settings page that calls cycleRememberDeviceToken().
> 
> Yep, good idea. I made those changes, and added success messages (following Grant's pattern) for Deactivate 2FA and Forget Trusted Devices.
> 
> Also, it seemed like remember token was only being saved if the user object happened to be saved elsewhere in the same request, so I made sure RememberDeviceTrait always saves the token to DB. But I'm confused about why it was working at all, previously, so please sanity check this @Khristophor [cd08029](https://github.com/GCTC-NTGC/TalentCloud/commit/cd080293fd8c3d44a04173f905d599c88e05022a)

You are correct, it did only seem to be saving when the user was also saved. I did some more digging and different traits/contracts seem to handle different portions of the update cycle for remember tokens for some reason:
https://github.com/laravel/framework/blob/6.x/src/Illuminate/Auth/EloquentUserProvider.php#L80-L98

That would be what we're missing, not sure where it should live in the functionality of our app, but it's probably okay in the trait for now.

app/Traits/RememberDeviceTrait.php
```diff
@@ -34,6 +34,8 @@ public function getRememberDeviceToken()
     public function setRememberDeviceToken($value)
     {
         if (! empty($this->getRememberDeviceTokenName())) {
+            // Make sure new value is immediately saved to db, as well as to model object.
+            $this->where('id', $this->id)->update([$this->getRememberDeviceTokenName() => $value]);
```
Khristophor: Can't we just call `$this->save()` after updating the attribute? Also looking at the Laravel implementation it looks like they prevent timestamps from updating during this process for some reason?
```php
        $user->setRememberToken($token);
        $timestamps = $user->timestamps;
        $user->timestamps = false;
        $user->save();
        $user->timestamps = $timestamps;
```
https://github.com/laravel/framework/blob/6.x/src/Illuminate/Auth/EloquentUserProvider.php#L80-L98

tristan-orourke: I suppose they want to keep the update_timestamp reflecting when the user purposefully updated their data.  That's sort of what I was trying to do here by only updating the one field.

tristan-orourke: I'll switch it do saving the same way as laravel does the remember token there. It'll be consistent, at least.

PR #1736, opened 2019-10-16
Bugfix - Copy to clipboard in IE

### LINKED ISSUE #1681: Bug - Copy to Clipboard buttons throw error in IE
# Description
Clicking the 'Copy to Clipboard' buttons on the JPB returns the following error in IE:
`SCRIPT438: Object doesn't support property or method 'remove'`
`JobBuilderRoot.js (1,751082)`

And the page scrolls to the bottom. The correct content does get copied to the clipboard though.

# Steps to Reproduce
1. Go to `/en/manager/jobs/<id>/builder/environment`
2. Click on 'Copy to Clipboard'
3. See error (with the inspector open)

yonikid15:
Resolves #1681 

### Notes:
- Refactored `copyToClipboard` to use methods supported in IE.


resources/assets/js/helpers/clipboard.ts
```diff
@@ -1,11 +1,30 @@
...
+    const textarea = document.createElement("textarea");
+    textarea.textContent = text;
+    textarea.style.position = "fixed"; // Prevent scrolling to bottom of page in MS Edge.
+    document.body.appendChild(textarea);
+    textarea.select();
+    try {
+      return document.execCommand("copy"); // Security exception may be thrown by some browsers.
+    } catch (ex) {
+      console.warn("Copy to clipboard failed.", ex);
+      return false;
```
tristan-orourke: Are you returning false to indicate failure? If so, its more clear to return `Promise.reject()` instead. Then you can change the method signature to `Promise<void>` instead of `Promise<void | boolean>`.

PR #1752, opened 2019-10-18
Add recovery codes which can be used as second auth factor.

### LINKED ISSUE #1664: Task - Add recovery code capability
## Description:

In order to allow users to access their account when changing devices, or reset their account when locked out, we should look into adding the [Recover Codes](https://github.com/antonioribeiro/recovery) package.

- [ ] This task is complex and requires acceptance criteria

tristan-orourke:
Resolves #1664.


Khristophor:
Looking good so far, made a couple changes:

1. Most account systems I've used will only show recovery codes once (when adding 2FA) and either never show them again, or allow you to generate another set. It says it's for security, but I can't find any literature explaining *why* this is secure.
2. I added a separate profile page fieldset for recovery codes, just to clean things up when some of the other branches are merged in.

PR #1817, opened 2019-10-25
2FA - OTP entry before 2FA activation

### LINKED ISSUE #1800: Feature - Require entry of a one time password before activating 2fa
# Description
This is to prevent accidental activation without completing setup in an app, locking you out of your account.

yonikid15:
Resolves #1800


yonikid15:
> I made some substantial changes. Continuing from the initial 2fa activation page requires a one time password on the same page. An incorrect or missing password stops activation.

Good solution. Moving the OTP to the same page as the 2fa would have helped avoid some problems I had to solve.

PR #1851, opened 2019-10-30
Task/job title placeholder

### LINKED ISSUE #1605: Feature - Manager - Placeholder Job title for non-translated job posters on dashboard
# Description

**Is your feature request related to a problem? Please describe.**
Currently if you complete a job poster in French, the title will be empty in English. 

**Describe the solution you'd like**
It would be better if we could provide a placeholder title "(French) Title" or similar. The link should also go to the page which has the info filled out.

Update: Add new field to Job Poster model which represents which language the poster was built in.

yonikid15:
Resolves #1605 
Dependent on #1929 

### Notes:
- "The link should also go to the page which has the info filled out." Still needs to be done. 
- Is there a way to determine if job poster was built in french or english? (#1929 resolves this)


app/Http/Controllers/JobController.php
```diff
@@ -67,12 +67,29 @@ public function managerIndex()
             ->withCount('submitted_applications')
             ->get();
 
+
+        /*
+            Check if job has a title. Append both bool (and french title if locale is 'en').
+            Because global fallback-locale set to true (config/translatable.php) any french translatables are set to the english copy. 'getTranslationsArray()' will retreive real value of translations.
+        */
+        $locale = app()->getLocale();
+        foreach ($jobs as &$job) {
```
tristan-orourke: What does the ampersand do here?

app/Http/Controllers/JobController.php
```diff
@@ -67,12 +67,29 @@ public function managerIndex()
...
+            $jobTranslations = $job->getTranslationsArray();
+            $hasTitle = !!$jobTranslations[$locale]['title'];
```
tristan-orourke: Why the double '!'?

app/Http/Controllers/JobController.php
```diff
@@ -67,12 +67,29 @@ public function managerIndex()
             ->withCount('submitted_applications')
             ->get();
 
+
```
tristan-orourke: Is there anywhere else we'll want to use this logic? (If there is, we can attach this logic to the JobPoster model instead of the controller. If not, this is fine.)

app/Http/Controllers/JobController.php
yonikid15: It's just a quick way of turning it into a boolean. This is all changed though going to push up soon.

app/Http/Controllers/JobController.php
yonikid15: Possibly. Currently, not sure where else we can use it.

app/Http/Controllers/JobController.php
Khristophor: https://www.php.net/manual/en/language.references.pass.php

app/Http/Controllers/JobController.php
tristan-orourke: So it was basically type casting to boolean? I'd have done a more explicit `(bool) $your_var` or used `array_key_exists()`.

yonikid15: I was using it just to test something, but yes I definitely should use something more explicit.

tristan-orourke:
I simplified how the alternate language job title was set, and used the chosen lang for the Edit Job link as well.

app/Http/Controllers/JobController.php
```diff
@@ -62,38 +62,19 @@ public function managerIndex()
...
-                } else {
-                    $job->title = $jobTranslations['en']['title'];
-                }
-
-                $job->trans_required = $job->title;
             }
+
+            // Always preview and edit in the chosen language
+            $job->preview_link = LaravelLocalization::getLocalizedURL($chosen_lang, route('manager.jobs.show', $job));
+            $job->edit_link = LaravelLocalization::getLocalizedURL($chosen_lang, route('manager.jobs.edit', $job));
```
yonikid15: This is what I was looking for!!! 👏

PR #1858, opened 2019-10-31
Bugfix - Manager profile copy

### LINKED ISSUE #1854: Copy -  Missing translation on manager profile when no answer is provided.
# Description
There is currently some missing French copy on the manager profile:

image

# Changes
 French should read:
Cette information n'a pas été fournie.

Also, could we look into why this wasn't caught be our translation unit tests?

yonikid15:
Resolves #1854 

### Notes:
- Also includes commit to fix return home link on login page for managers.

### DIFF
--- a/app/Http/Controllers/ManagerProfileController.php
+++ b/app/Http/Controllers/ManagerProfileController.php
@@ -65,15 +65,13 @@ class ManagerProfileController extends Controller
             ]
         ];
 
-        $noInfo = $manager_profile['no_info'];
-
         return view('applicant/manager', [
             'manager_profile' => $manager_profile,
             'urls' => Lang::get('common/urls'),
             'manager' => $manager,
             'manager_profile_photo_url' => '/images/user.png', // TODO get real photo.
             'manager_profile_sections' => $manager_profile_sections,
-            'noInfo' => $noInfo,
+            'noInfo' => $manager_profile['no_info'],
         ]);
     }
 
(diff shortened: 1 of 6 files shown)


app/Http/Controllers/ManagerProfileController.php
```diff
@@ -65,13 +65,15 @@ public function show(Request $request, Manager $manager)
...
 
+        $noInfo = $manager_profile['no_info'];
 
         return view('applicant/manager', [
             'manager_profile' => $manager_profile,
             'urls' => Lang::get('common/urls'),
             'manager' => $manager,
             'manager_profile_photo_url' => '/images/user.png', // TODO get real photo.
             'manager_profile_sections' => $manager_profile_sections,
+            'noInfo' => $noInfo,
```
Khristophor: In the future I'd just 
```php
'no_info' => $manager_profile['no_info'],
```
but it's not hurting anything with the extra variable declaration

resources/views/auth/login.html.twig
```diff
@@ -15,7 +15,7 @@
         <section class="tc-auth" data-c-padding="tb(triple)" data-c-overlay="black(80)" style="background-image: url('/images/bg_crowd.jpg');">
             <div class="tc-auth-wrapper" data-c-padding="tb(triple)" data-c-container="small">
                 {# Return to the homepage link. #}
-                <a data-c-color="white" href="/" title="Return Home">
+                <a data-c-color="white" href={{ home_url }} title="Return Home">
```
Khristophor: Might as well localize the `title` while you're in here 😉

yonikid15: Missed that 👍

app/Http/Controllers/ManagerProfileController.php
yonikid15: Okay will do

Issue #1850, opened 2019-10-30
Feature - JPB - Warn user if they attempt to follow a link and have unsaved data on the page 


tristan-orourke:
# Description

Clicking on a link (at a minimum, links in the Progress Bar) should first warn user if they have unsaved data.

We currently do this in at least some parts of the Timeline application. This issue is about bringing them over to the JPB

gobyrne:
This is being done in some places, but not consistently. JPB and Application are the most relevant.

PR #1888, opened 2019-11-01
JPB - reverse clickable progress tracker

### LINKED ISSUE #1832: Task - Manager - Job Poster Builder: Make Progress Bar Clickable in Reverse
# Description

Right now, navigating the job poster builder is difficult and confusing because the progress bar doesn't allow for clicking between steps.

In an ideal world, it would work as follows:
- steps that have already been completed would become links
- once the user reaches the review step, all steps should be linked (including review, despite the fact that it is technically incomplete)
- hovering an unreachable step should provide a tool tip letting the user know they have to complete previous steps first

# Images

Example where the manager has only reached step 3:
image

Example when all the steps are complete:
image

yonikid15:
Resolves #1832 

### Notes:
- Meant to make a draft PR.
- Still needs approval of copy, and the french translation.


resources/assets/js/components/JobBuilder/JobBuilderStep.tsx
```diff
@@ -133,14 +133,17 @@ const JobBuilderStep: React.FunctionComponent<JobBuilderStepProps> = ({
...
     <section>
-      <JobBuilderProgressTracker
-        job={job}
-        tasks={keyTasks}
-        maxTasksCount={VALID_COUNT}
-        criteria={criteria}
-        dataIsLoading={dataIsLoading}
-        currentPage={currentPage}
-      />
+      {jobId !== null && (
```
tristan-orourke: This stops the progress bar from appearing on the first step, when you're creating the poster.

resources/assets/js/components/ProgressTracker/ProgressTrackerItem.tsx
```diff
@@ -37,10 +49,35 @@ const ProgressTrackerItem: React.FunctionComponent<
...
+          onClick={(e): void => {
+            e.preventDefault();
+            if (
+              window.confirm(
+                intl.formatMessage(jobBuilderMessages.unsavedChangesWarning),
+              )
+            ) {
+              navigate(link);
+            }
+          }}
```
tristan-orourke: Don't worry about throwing that alert here. We only want to show the alert when there is unsaved data, which means the form components themselves will have to intercept the navigation somehow.

p.s. We're trying to keep this component generic. If we kept this alert here, the message should be moved to somewhere other than jobBuilderMessages.

resources/assets/js/components/JobBuilder/JobBuilderStep.tsx
tristan-orourke: fixed

resources/assets/js/components/ProgressTracker/ProgressTrackerItem.tsx
tristan-orourke: I removed the onClick handler. We can make a better solution while working on #1850.

PR #1889, opened 2019-11-01
Returning to JPB brings you to last step visited, instead of Review page

### LINKED ISSUE #1833: Feature - JPB - Returning to JPB brings you to last step visited, instead of Review page
# Description
Leaving the job poster builder and subsequently returning should return you to the last step you visited, as opposed to the review page.
This is integral because right now the user is taken to a review page they might not have ever seen before, and also gives them link access to steps they haven't reached yet.

The JobBuilderProgressTracker component already checks which steps are completed or unvisited, to display the right icon in the progress tracker. Some of that code may be reusable here, to determine the last incomplete step (which you should be directed to).

tristan-orourke:
Resolves #1833.

### Notes:
- I created a new route, `jobs/{jobId}/builder`, which is routed (in the the JPB SPA) to a component which loads all the job data and (spa-style) redirects to the earliest incomplete step. 
- If all the steps are complete, it always redirects to the review.
- The "Submit" button on the manager job index still always goes to the review page, regardless of whether the job is complete or not.
- ...It occurs to me now that putting the redirect logic on the backend might be a better user experience, because it wouldn't have to wait for the data to load, but the logic already existed on the frontend, so this was the fastest way to build it.


resources/assets/js/components/JobBuilder/jobBuilderHelpers.ts
```diff
@@ -237,3 +246,31 @@ export const isJobBuilderComplete = (
...
+    case "impact":
+      return jobBuilderImpact;
+    case "tasks":
+      return jobBuilderTasks;
+    case "skills":
+      return jobBuilderSkills;
+    case "review":
+      return jobBuilderReview;
+  }
+};
```
yonikid15: `pageToUrl` is missing a return type `((locale: string, jobId?: number | undefined) => string)`
Eslint is complaining that switch case doesn't have a `default`. Maybe you could set the default `jobBuilderReview`?

resources/assets/js/components/JobBuilder/RedirectToLastIncompleteStep.tsx
```diff
@@ -0,0 +1,169 @@
...
+  jobId: number;
+  job: Job | null;
+  jobIsLoading: boolean;
+  tasks: JobPosterKeyTask[];
+  maxTasksCount: number;
+  tasksIsLoading: boolean;
+  criteria: Criteria[];
+  criteriaIsLoading: boolean;
+  redirect: (url: string) => void;
+}
```
yonikid15: Name `redirect` is already declared. Maybe try changing the prop to `redirectToUrl` to avoid confusion.

tristan-orourke: I went with handleRedirect

PR #1915, opened 2019-11-05
Add device memory to 2FA flow

### LINKED ISSUE #1663: Task - Prompt user for one time password
## Description:

Currently the one time password is requested on every login, we need to add a field to the User table to keep track of the last successful OTP entry and calculate when to prompt again. Other sites use a checkbox or similar with the text "Don't ask again for another x days", and possibly detect a new device to prompt again.

Decide on the scope (simple time check vs multi-device memory) and implement.

- [ ]  This task requires unit testing

Khristophor:
Resolves #1663

### Notes:
- Saves a cookie locally after successful OTP entry if "remember device" is checked
- Cookie expiration is set for one month (handled by the browser)
- Checks for cookie presence during OTP middleware
- Need to add some testing


app/Http/Controllers/Auth/TwoFactorController.php
```diff
@@ -99,10 +102,24 @@ public function confirm(Request $request)
...
         return redirect($expectedUrl);
     }
+
+    protected function rememberDevice(Request $request)
+    {
+        $remember = $request->input('remember_device');
+        if ($remember) {
+            Cookie::queue(
+                'remember_device',
+                'yes',
```
Khristophor: Not entirely sure what to use as the content for the cookie. It's hashed/encrypted client side, and we pretty much just need to check for it's existence. We could do something similar to the "remember password" like storing a token in the user table and validating against that.

Khristophor:
I looked through the controller and I'm not entirely sure how to unit test this. There are only a couple of routes and they all require OTP entry for any valuable testing. I could try splitting it up and importing the controller directly but that seems weird.

app/Http/Controllers/Auth/TwoFactorController.php
tristan-orourke: I think we should do the same as the "remember password" token. Right now its not user specific - once I've checked "remember this device" as one user, I can log in as another user and still skip 2fa.

tristan-orourke: I've been looking at how Laravel uses that remember password token/cookie (in `vendor\laravel\framework\src\Illuminate\Auth\SessionGuard.php`). It cycles the token regularly (on every logout) which we should consider as well. It would help stop someone from stealing the cookie and using it indefinitely. 

Some other thoughts:
- There's nothing stopping someone from copying the cookie onto another device.
- Using the cookie like this would mean only one user can be remembered per device (probably fine).

app/Http/Middleware/Google2FA.php
```diff
@@ -15,10 +17,28 @@ class Google2FA
...
+            $remember = $request->cookie($user->getRememberDeviceKey());
+
+            // If tokens do not match, cookie is no longer valid.
+            if ($remember !== null && $remember !== $user->getRememberDeviceToken()) {
+                Cookie::forget($user->getRememberDeviceKey());
+            }
+
+            if ($authenticator->isAuthenticated() || ($remember !== null && $remember === $user->getRememberDeviceToken())) {
+                if (!$authenticator->isAuthenticated()) {
+                    Log::notice('User skipped OTP entry with known device.', ['id' => $request->user()->id]);
```
tristan-orourke: I think we should call $authenticator->login() at this point. Otherwise, this message will be logged with every single request, and we'll have to add these same token checks to our Require2FA middleware.

Khristophor: Ah good point, I'll do that

PR #1933, opened 2019-11-06
Add new classification code

gggrant:
Resolves blocker for Rosita making a job poster.

### Notes:
- Includes some other misc language fixes.
- The classification code is AD, which is the same as AS but for a different department.
- Yeah.

### DIFF
--- a/resources/assets/js/components/JobDetails/JobDetails.tsx
+++ b/resources/assets/js/components/JobDetails/JobDetails.tsx
@@ -31,6 +31,7 @@ import {
   FrequencyId,
   TravelRequirementId,
   OvertimeRequirementId,
+  ClassificationId,
 } from "../../models/lookupConstants";
 import { emptyJob } from "../../models/jobUtil";
 import {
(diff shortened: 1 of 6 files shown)


resources/assets/js/components/JobDetails/JobDetails.tsx
```diff
@@ -333,20 +334,7 @@ const JobDetails: React.FunctionComponent<
...
-          "CS",
-          "EC",
-          "EX",
-          "FO",
-          "IS",
-          "PC",
-          "PE",
-          "PM",
-        ],
+        Object.values(ClassificationId),
```
tristan-orourke: To keep the rest of the code equivalent, you actually want `Object.keys(ClassficiationId)` here, not Object.values

tristan-orourke: Fixed in 33cbd558a98e856a84c68cc61da1d5055328f4b4

tristan-orourke:
NOTE: For this new classification to be stored correctly, it also needs to be added to the database. That can be done on the Admin portal.

Issue #372, opened 2018-10-17
Applicant - Relative modals on job application


gobyrne:
# Description
Tie in the modals that Josh made for the Job application so that applicants can add references and work samples to their skills directly on the job application.

gobyrne:
I think it's a duplicate of #463

tristan-orourke:
Actually, this isn't complete, and its also not a duplicate.

gobyrne:
Can you please elaborate here on in the description?

gobyrne:
Waiting on front-end framework (#701)

Khristophor:
Created the Modal in React as a component, but in order to properly implement it I think we should do some heavy refactoring of the relevant applicant sections (Application Skills and Applicant Profile). This will involve adding API routes, creating the required React components for the containers, creating a Skill component and retrieving the data.

yonikid15:
My bad. Accidentally closed the issue.

PR #1938, opened 2019-11-06
Fixes 'Submit Feedback' link and various accessibility issues

### LINKED ISSUE #1448: Bug - "Submit Feedback" link does work
# Description
This bug was submitted by a user: The “provide Feedback” link doesn’t work. I had to right-click “copy email address” and paste it into outlook to get it to work.

# Steps to Reproduce
I tried to reproduce the bug on both IE and Chrome, on both the applicant and the manager portal, but it auto-generated an empty email for me after I clicked the link.

# Images
Place image attachments here.

gggrant:
Resolves #1448.


resources/views/applicant/application_post_02.html.twig
```diff
@@ -35,6 +35,6 @@
 
     {# Modals #}
 
-        {% include "common/modal" with {"page":application_template} %}
+        {% include "common/modal" with {"page":application} %}
```
Khristophor: Is this modal actually used for anything? I'm not seeing a way to open it locally.

gggrant: #372

One day, maybe.

PR #1960, opened 2019-11-08
Immutable Application Data

### LINKED ISSUE #647: Task - Make Application data immutable
# Description

Data submitted in an application should be separated from Applicant profile, which can change.

The skills and experience submitted in an application are just links to the data on an applicant's profile. If they change their profile later, there will be no record of the data originally submitted. 

Hiring managers should only have access to the skills and experience that was submitted with an application, not the full info from applicants profile.

tristan-orourke:
Resolves #647


tristan-orourke:
Oh! We need to make sure no routes are set to require 2fa for now.

PR #2047, opened 2019-11-14
Claimed jobs API routes for HR advisors

### LINKED ISSUE #1894: Task - HR Portal - Dashboard - Create API endpoints allowing hr advisers to 'claim' and unclaim jobs
# Description
HR advisers need to be able to 'claim' jobs. A single adviser can claim many jobs, each job can have multiple claimants.
This will require changes to the DB, a new HR Advisor data model, as well as a new controller.

# Acceptance criteria
- [x] data migrations
- [x] functioning api routes
- [x] phpunit tests for api routes

gggrant:
Resolves #1894.


app/Http/Controllers/Api/HrAdvisorController.php
```diff
@@ -0,0 +1,97 @@
...
+     * Update the specified resource in storage.
+     *
+     * @param  \App\Models\HrAdvisor  $hrAdvisor
+     * @param  \App\Models\JobPoster  $jobPoster
+     * @return \Illuminate\Http\Response
+     */
+    public function update(HrAdvisor $hrAdvisor, JobPoster $jobPoster)
+    {
+        $hrAdvisor->claimed_jobs()->attach($jobPoster->id);
+        return response()->json();
```
Khristophor: I don't think lines like this will actually return anything, which might be fine for now, but might want to just return a `response->json(['status' => 'ok']);` or something for the sake of testing.

app/Http/Controllers/Api/HrAdvisorController.php
```diff
@@ -0,0 +1,97 @@
...
+     * @param  \App\Models\HrAdvisor  $hrAdvisor
+     * @param  \App\Models\JobPoster  $jobPoster
+     * @return \Illuminate\Http\Response
+     */
+    public function claimJob(HrAdvisor $hrAdvisor, JobPoster $jobPoster)
+    {
+        // $hrAdvisor = HrAdvisor::find(1);
+        $jobPoster = JobPoster::find($hrAdvisor->id);
+        $hrAdvisor->claimed_jobs()->attach($jobPoster);
+        return response()->json();
```
Khristophor: Same thing here, you'd probably want to return the result of the `attach()` (assuming it returns something) for success or `false` for a failure case, which can be passed along to the browser in some way in the `update` call.

routes/web.php
```diff
@@ -510,4 +544,11 @@ function (): void {
     // User must be logged in to user currentuser routes
     Route::get('currentuser/manager', 'Api\ManagerApiController@showAuthenticated')
         ->middleware('auth');
+
+    // HR Portal - Claims
+    Route::resource('hr', 'HrAdvisorController');
```
Khristophor: I think this needs to come out, it's trying to generate routes to methods in `App\Http\Controllers\HrAdvisorController.php`

PR #2072, opened 2019-11-18
Use SPA routing when linking from JPB review page to eariler steps.

### LINKED ISSUE #1353: Bug - Job Poster Builder - some links between steps still not using the navigate() function
# Description
Some links between JPB steps, including the links on the Review page, are still making normal requests, not SPA-style navigation. Switching to SPA-style requests will significantly speed up page transitions.

tristan-orourke:
Resolves #1353.

### DIFF
--- a/resources/assets/js/components/JobBuilder/Review/JobReview.tsx
+++ b/resources/assets/js/components/JobBuilder/Review/JobReview.tsx
@@ -202,11 +202,9 @@ const JobReviewSection: React.FunctionComponent<JobReviewSectionProps> = ({
             data-c-grid-item="tp(1of2)"
             data-c-alignment="base(centre) tp(right)"
           >
-            <Link url={link} title={linkLabel}>
-              <>
-                <i data-c-colour="c2" className="fas fa-edit" />
-                {linkLabel}
-              </>
+            <Link href={link} title={linkLabel}>
+              <i data-c-colour="c2" className="fas fa-edit" />
+              {linkLabel}
             </Link>
           </div>
         </div>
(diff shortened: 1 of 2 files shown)


resources/assets/js/components/JobBuilder/Review/JobReview.tsx
```diff
@@ -202,10 +202,10 @@ const JobReviewSection: React.FunctionComponent<JobReviewSectionProps> = ({
             data-c-grid-item="tp(1of2)"
             data-c-alignment="base(centre) tp(right)"
           >
-            <a href={link}>
+            <Link href={link} title={linkLabel}>
               <i data-c-colour="c2" className="fas fa-edit" />
```
Khristophor: @tristan-orourke just wanted to point out my change here, fragments are only required if you're returning sibling tags from a component. You can pass any amount of stuff (even as siblings) as children, and as long as the consuming component only returns a single tag it's fine.

resources/assets/js/helpers/router.tsx
```diff
@@ -70,3 +70,20 @@ export const navigate = (url: string): void => {
 export const redirect = (url: string): void => {
   HISTORY.replace(url);
 };
+
+export const Link: React.FC<{ href: string; title: string }> = ({
+  href,
+  title,
+  children,
+}): React.ReactElement => (
```
Khristophor: @tristan-orourke couple changes here as well. Our linter was complaining about the `url` prop, saying the anchor did not have a valid `href` so I just used the same variable for both. I also condensed the return, if you're just returning a fragment/tag, you can forego the `return` keyword and just return the tag directly.

resources/assets/js/components/JobBuilder/Review/JobReview.tsx
tristan-orourke: Awesome, thanks.

PR #2075, opened 2019-11-18
Jest code coverage / JUnit support / config tweaks

gggrant:
Started off as adding JUnit output for PHPUnit and Jest.

Also adds Jest coverage to CI build, which is sent to Codecov.

Scrutinizer runs on JS files as well now (was blocked previously with `resources/` on the exclude list.


gggrant:
Edits to `package.json`:

```
"jest-junit": "^9.0.0",
"jest-junit": {
        "outputDirectory": "./report/",
        "outputName": "junit.jest.xml"
    },
```

PR #2085, opened 2019-11-19
Task - viewing application data

### LINKED ISSUE #1700: Task - When viewing a submitted JobApplication, show saved copies of Applicant data.
Depends on #1696.

When viewing a completed (ie submitted) JobApplication, saved copies of Applicant data should be shown, instead of data directly from the Applicant profile. This goes for:

- [ ] degrees
- [ ] courses
- [ ] work_experience
- [ ] skill_declarations
- [ ] references
- [ ] work_samples

Note: when a user is previewing a not-yet-submitted application, it should still display data directly from the applicant profile.

yonikid15:
Resolves #1700 

### Notes:
- Created a `isDraft` method which returns bool.
- Now when viewing a job application with `draft` status it should grab data through the Applicant model. Otherwise, it will grab it from a copy of the applicant data (which is currently getting done by Tristan).

### DIFF
--- a/app/Http/Controllers/ApplicationController.php
+++ b/app/Http/Controllers/ApplicationController.php
@@ -58,6 +58,21 @@ class ApplicationController extends Controller
             $application->load(['veteran_status', 'citizenship_declaration', 'application_review', 'applicant.user']);
         }
 
+
+        // If the application status is draft then get data through the applicant model. Else, grab the data from the application itself.
+        if (!$application->isDraft()) {
+            $source = $application->applicant;
+        } else {
+            $source = $application;
+        }
+
+        $degrees = $source->degrees;
+        $courses = $source->courses;
+        $work_experience = $source->work_experience;
+        $skill_declarations = $source->skill_declarations;
+        $references = $source->references;
+        $work_samples = $source->work_samples;
+
         return view(
             $view,
             [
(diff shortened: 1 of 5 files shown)


app/Http/Controllers/ApplicationController.php
```diff
@@ -58,6 +58,21 @@ public function show(JobApplication $application)
             $application->load(['veteran_status', 'citizenship_declaration', 'application_review', 'applicant.user']);
         }
 
+
+        // If the application status is draft then get data through the applicant model. Else, grab the data from the application itself.
+        if (!$application->isDraft()) {
```
tristan-orourke: Oops, this check is backwards :P We want to go through the applicant if it IS a draft.

app/Models/JobApplication.php
```diff
@@ -267,4 +267,19 @@ public function getMeetsEssentialCriteriaAttribute():bool
...
+
+
+    /**
+     * Check if the status of the application is 'draft'
+     *
+     * @return boolean
+     */
+    public function isDraft() : bool
+    {
+        if ($this->application_status_id == '1') {
```
tristan-orourke: Using the application_status name would make the code more clear.

app/Http/Controllers/ApplicationController.php
yonikid15: Oops, forgot to switch that back :p I was testing it out.

PR #2097, opened 2019-11-20
Task - Unclaimed job grid

### LINKED ISSUE #1936: Task - HR Portal - Dashboard job grid
Subtask of #1533

# Wireframes / Images

Prototype Link: https://joshbeveridge.ca/tc-manager-dash/hr-adviser/

yonikid15:
Resolves #1936 

### Notes:
- Added a grid component, which takes a list of `UnclaimedJob` objects.
- The grid component can be viewed in Storybook under "Unclaimed Job Card" story.


yonikid15:
> Everything's good, good-wise. But now that we can see how simple UnclaimedJobGrid is, I wonder if its worth giving it its own component. Given that its just a map function inside a single div, I think it might be clearer to include that directly in whatever parent component this eventually ends up in. @yonikid15 Thoughts?
> 
> edit: here's how it could look https://github.com/GCTC-NTGC/TalentCloud/tree/task/unclaimed-job-grid-componentless

I think it's a matter of preference at this point. I was thinking maybe we can extend/build the use of the `UncliamedJobGrid` component in the future if needed. Or if the shape of the grid changes in the future we only need to change the logic in one place (if we end up using it more then once). However, right now we can use your branch instead and change it later if need be.

PR #2123, opened 2019-11-21
Bugfix - SPB - remove word limit on questions

### LINKED ISSUE #1603: Bug - Screening plan builder - word limit for questions
# Description
When typing a longer-worded question into the under rating guide builder of the Assessment Plan Builder, an error message will pop up. It's likely related to word limit, but that limit is not specified.

As the error message automatically disappears after it’s displayed the manage couldn't recall which fields had given the error if they are working on multiple questions.

# Steps to Reproduce
1. Go to screening plan builder then rating guide builder
2. Add a question and start typing
3. The error message will eventually appear when the question gets long enough

# Images

image

yonikid15:
Resolves #1603 

### Notes:
- I changed the datatype of the field from `string()` to `text()`. 
- Also, allowed user to set the input field to empty.


resources/assets/js/components/UpdatingInput.tsx
```diff
@@ -76,7 +73,9 @@ class UpdatingInput extends Component<UpdatingInputProps, UpdatingInputState> {
             }
           }
         }}
-        onBlur={handleSave}
+        onBlur={(event: React.ChangeEvent<HTMLInputElement>) => {
+          if (event.target.value.length > 0) handleSave();
```
tristan-orourke: There are cases where we may want to allow 0-length answers, so I don't think we can skip saving in this case.

It may be worth trying to check to see if the input has changed since the last save? That's probably complex enough to make a separate PR though.

PR #2131, opened 2019-11-22
Bugfix - JPB - Total number of tasks

### LINKED ISSUE #2096: Bug - Inconsistent number of tasks allowed on JPB
# Description
On the job poster builder we currently limit the number of task you can have to 8. There is text at the top of the page that says you should aim for 4-6 but we allow no more than eight. Finally there is error text that appears when you have more than 8 which says you must have no more than 6.

These should all be reduced to 6. So the actual limit should be six and the text at the top of the page should say, aim for 4-6 but will be limited to 6.

# Steps to Reproduce
1. Go to Tasks page of JPB
2. Click on add task until you have more than 8
3. Compare error text to other text on the page

yonikid15:
Resolves #2096 

### Notes:
- Changed the name of the constant to make more explicit. 
    - `VALID_COUNT` -> `NUM_OF_TASKS`
- Instead of importing the constant into each file using `isJobTasksUntouched`. I removed the argument and called the global value instead. 
- Updated the copy.

### DIFF
--- a/resources/assets/js/components/JobBuilder/JobBuilderStep.tsx
+++ b/resources/assets/js/components/JobBuilder/JobBuilderStep.tsx
@@ -9,7 +9,7 @@ import {
   Skill,
   Department,
 } from "../../models/types";
-import { VALID_COUNT, JobBuilderPage } from "./jobBuilderHelpers";
+import { NUM_OF_TASKS, JobBuilderPage } from "./jobBuilderHelpers";
 import {
   getJob,
   getTasksByJob,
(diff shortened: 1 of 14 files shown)


resources/assets/js/components/JobBuilder/JobBuilderStep.tsx
```diff
@@ -9,7 +9,7 @@ import {
   Skill,
   Department,
 } from "../../models/types";
-import { VALID_COUNT, JobBuilderPage } from "./jobBuilderHelpers";
+import { NUM_OF_TASKS, JobBuilderPage } from "./jobBuilderHelpers";
```
tristan-orourke: This is one unused import you missed deleting, I took care of it.

PR #2195, opened 2019-11-29
Replace classification accessor on Job Posters with 1-n relationship

### LINKED ISSUE #1445: Bug - Admin - Job Poster classification accessor performance
# Description
The recently added classification accessor on the Job Poster model causes some pretty heavy querying whenever the property is called in a template.

This function:
```php
    public function getClassificationCodeAttribute()
    {
        if ($this->classification_id !== null) {
            $classification = Classification::find($this->classification_id);
            return $classification->key;
        }
        return null;
    }
```
is called every time, which for about 10 Job Posters on a page resulted in over 200 extra queries to be run.

The classification property should be removed from the Job Poster model and replaced with a relationship to the actual classification table.
**NOTE:** Templates need to be adjusted accordingly after this change.

# Steps to Reproduce _with debugbar enabled_
1. Ensure you have some Job Posters seeded
2. Navigate to a page with Jobs that display the classification (`/manager/jobs`)
3. Look at the "queries" section
4. Check out all the duplicate classification queries

gggrant:
Resolves #1445.


yonikid15:
@gggrant  The problem was the the previous field `classification_code` was holding the value of the string value (eg. AS). With the new changes you made, `classification_id` holds the id (int). However, the data on the react side is expecting the string value, so I had to refactor a few things.

app/Http/Controllers/JobController.php
```diff
@@ -57,21 +57,20 @@ public function managerIndex()
         $manager = Auth::user()->manager;
 
         $jobs = JobPoster::where('manager_id', $manager->id)
+            ->with('classification')
```
Khristophor: This was the key piece missing to actually fix the issue I reported 😆

yonikid15: 😢. This is the gold right here eh: https://laravel.com/docs/5.8/eloquent-relationships#eager-loading

resources/views/applicant/jpb_job_post/basics.html.twig
```diff
@@ -4,104 +4,64 @@
...
+				{{ job_post.basics.security_label }}
+			</span>
+			<span class="job-poster__basics-content">
+				{{ job.security_clearance_id ? job.security_clearance.value : noInfo }}
+			</span>
+			<span class="job-poster__basics-label">
+				{{ job_post.basics.classification_label }}
+			</span>
+			<span class="job-poster__basics-content">
+				{{ job.classification.key }}-{{ job.classification_level ?: noInfo }}
```
Khristophor: Looks like this will display the key and the noInfo fallback if the level is missing? Also, I'm not sure about display requirements, but the level is usually left padded with a `0` if applicable.

I'd rather do null checking and massage the output in the controller in this case so the frontend just receives the nicely formatted content.

resources/views/manager/job_index/job.html.twig
```diff
@@ -6,153 +6,131 @@
...
+					{{ humanizeDate(job.review_requested_at) }}</span>
+			{% endif %}
+			{% if job.published_at %}
+				<span class="poster-date-time--published">{{ job.classification.key }}-{{ job.classification_level }} - {{ jobs_l10n.published_on }}
+					{{ humanizeDate(job.published_at) }}</span>
+			{% endif %}
+			{% if job.close_date_time %}
+				<span class="poster-date-time--closed">{{ job.classification.key }}-{{ job.classification_level }} - {{ jobs_l10n.closed_on }}
+					{{ humanizeDate(job.close_date_time) }}</span>
+			{% endif %}
```
Khristophor: Same here with `classification_level`, we should check if it should be left padded with a zero. Display on this should be consistent across the site if possible, probably with a mutator on the Classification model itself.

resources/assets/js/components/JobBuilder/Details/JobDetails.tsx
```diff
@@ -1098,7 +1025,10 @@ const JobDetails: React.FunctionComponent<
                               securityClearance(Number(values.securityLevel)),
                             )
                       }
-                      classification={String(values.classification)}
+                      classification={getKeyByValue(
+                        ClassificationId,
+                        values.classification,
+                      )}
                       level={String(values.level)}
```
Khristophor: This needs a pad too now that I think about it

resources/assets/js/components/JobBuilder/Details/JobDetails.tsx
```diff
@@ -176,17 +174,20 @@ interface DetailsFormValues {
...
-  classification: string,
+  classification: number | string,
   intl: IntlShape,
 ): string => {
-  return hasKey(educationMessages, classification)
-    ? intl.formatMessage(educationMessages[classification])
-    : "";
+  return hasKey(educationMessages, classificationCode(classification))
+    ? intl.formatMessage(educationMessages[classificationCode(classification)])
+    : "he";
```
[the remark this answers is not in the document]
yonikid15: Forgot to remove that.

resources/views/applicant/jpb_job_post/basics.html.twig
yonikid15: I think null checking in the controller should be put into another issue, since it's being done in multiple places. In another issue we can remove all of them together. Sound good @Khristophor ?

Khristophor: Sure, as long as we don't have a classification code without a level

PR #2276, opened 2019-12-06
HR Portal - Job Index page

### LINKED ISSUE #1533: [Design] - HR: Job Index
Design Goal
**Please explain the primary goals of this design work.**

The purpose of this feature is to allow HR Advisors to quickly and easily get to the information that they are looking for and to respond quickly to issues that require their review, input or approval.

# Research
**Please list/link to all research performed for or referenced in this design work. This can include research you've done in the past. Include research relevant to insights from team members as well. Don't be afraid to ask for references or documents! This section should also incorporate data collected from user testing.**

- See HR Advisor Portal / Commenting #1532 

## Discussion w/ @laurenoncoffee, @gobyrne, @nicolsonaitken , @joshdrink , and @JasmitaChawla 
- it would be beneficial to include dynamic filters for:
  - manager
  - classification
  - has vs. does not have PSC process # 
- jobs should be classified into 2 main groups: active vs. complete
  - Lauren suggested this based on previous discussions with HR and her knowledge of of the archive structure
- jobs should display whether they have a PSC job process number, priority clearance number, and a letter of offer date issuance (entered via concierge)
- jobs should display both the time it has been with a manager and the time it has been with HR
- jobs should only be able to be punted back to managers if the adviser has either:
  - commented
  - submitted it for approval
    - can this be taken back?

@nicolsonaitken **Prototype Deliverable: an HR view of all jobs assigned to them, each with a chronological history of comments associated to that job action, as well as the "send back to manager" and "final approval" buttons.**

## Discussion w/ @laurenoncoffee @gobyrne @nicolsonaitken @Jerryescandon @JasmitaChawla @joshdrink 
- HR advisers will be able to claim a specific job through the job summary page (see issue #1763 ) 
  - this should put claimed jobs into a defined section at the top of the dashboard
- HR advisers need something new and polished to interact with, so we should focus on making this UI slightly more visual

## Assumptions
**Please list all assumptions made during your design work. This should include behavioural assumptions made by the team that have no direct research references.**

- HR Advisors will want to get to the relevant information as quickly as possible as noted within the HR Super User Workshops held on September 25th.
- per discussion with @nicolsonaitken we're assuming that HR Advisers have enough information on the card to know whether they should claim the job

# Important Notes / Copy
**This is a space to include important notes for other team members, including things like key copy, questions that need to be asked at a later date, etc.**

Job Index Status States include:
- Draft
- In Review
- In Translation
- In Approval
- Approved
- Ready to Post
- Published
- Closed
- Complete

## Chess Clock Feature Estimate
@tristan-orourke and I discussed the back'n'forth feature and he feels it'll be 3 days worth of pure development time, assuming the following:
- the commenting feature is already built, and built with this in mind
- we're not accounting for dev deployment/testing/bugfixing

# Wireframes / Images

Prototype Link: https://joshbeveridge.ca/tc-manager-dash/hr-adviser/

## Version 1

**Dashboard at a Glance**
- Posters Needing Your Attention
- New Posters Not Yet Completed

image

**Wireframes**
- [Dashboard V1](https://xd.adobe.com/view/e5416bde-30e3-49bc-5790-5e77d8a4d22c-9370/?fullscreen)

## Version 2
- adds context
- breaks page up by claimed vs unclaimed
- brings back card format for unclaimed
- matches claimed jobs to the manager view

manager-dashboard-sketch

## Version 3
- [Dashboard V3](https://xd.adobe.com/view/6d438863-6505-4807-5996-ddab663b1349-1b72/?fullscreen)

## Version 4

Link: https://joshbeveridge.ca/tc-manager-dash/hr-adviser

- added activity link and volume
  - this will work based on login timestamp - if there has been new activity since the last time you logged in, add it to the count
- hid the filters until they're needed

## Version 5

- added hiring manager (this is obvious needed info, it was just forgotten)
- add Talent Cloud specific timer
  - this was added as a result of @gobyrne and @nicolsonaitken because eventually the job action hits translation/concierge, and we should have a timer for this

image

### Future Features
- manager timer
- hr timer
- filters for the department section
- load more/pagination for the department section
- eventually we'll likely need filters for completed jobs too

**Additional Feature Requested**
During the last design review @laurenoncoffee requested the addition of a clock icon next to the Time with you, Time with Manager, Time with Talent Cloud as an addition visual cue or nudge to be added to the next version.

## Version 6

Prototype link: https://joshbeveridge.ca/tc-manager-dash/hr-adviser

### Changes
- updated the timer to reflect the new requests found in #2122 

### Screenshot
image

### LINKED ISSUE #2349: Task - HR - Finish UI of Job Index page
# Description
Complete the UI of the HR Job Index page. The complete page should be viewable in storybook, but it doesn't need to communicate with the server.

See design issue #1533 for details.

yonikid15:
UPDATE: Depends on #2311 
Resolves #2349

### Notes:
- Created Job Index page component (JobIndexPage.tsx).
- Can be viewed in Storybook under "HR Portal" section.
- Snapshot test


tristan-orourke:
@yonikid15 Gray and I decided to use Design issues like #1533 as mini Epics instead of treating them like normal dev stories, so I disconnected this PR from #1533 and made a simpler dev story that more succinctly describes what you did here.

resources/assets/js/models/types.ts
```diff
@@ -248,3 +248,37 @@ export interface JobPosterKeyTask {
...
+  userTime: number;
+}
+
+export interface UnclaimedJob {
+  title: string;
+  createdAt: string;
+  status: JobStatus;
+  hiringManagers: string[];
+  hrAdvisors: string[];
+}
```
tristan-orourke: This file is meant to contain Domain types, the most basic representations of the data the app uses. These interfaces look like they're pretty specific to individual components. They should just be exported from those component files, if necessary.

Activity, and things like managerTime/userTime, will eventually need domain types, but I think we're still a bit too unclear on how they should work to add them yet. 😅

resources/assets/js/components/HRPortal/JobIndexPage.tsx
```diff
@@ -0,0 +1,213 @@
...
+                using the 'Remove Myself From This Job' button."
+            />
+          </p>
+
+          {/* Unclaimed Jobs List */}
+          <section data-c-grid="gutter">
+            {unclaimedJobs.length !== 0 &&
+              unclaimedJobs.map(
+                (unclaimedJob): React.ReactElement => {
+                  return <UnclaimedJobCard {...unclaimedJob} />;
```
tristan-orourke: UnclaimedJobCard should also receive a url. Right now the component title is always a url with blank href.

tristan-orourke:
@joshdrink  I changed the Unclaimed Job Card a bit, I'd like your review on that. Mostly I just made "Claim This Job" a button instead of a link.

resources/assets/js/components/HRPortal/JobIndexPage.tsx
```diff
@@ -0,0 +1,216 @@
...
+import { JobStatus } from "../../models/lookupConstants";
+
+interface CompletedJobsAccordionProps {
+  completedJobActions: JobCardProps[];
+}
+
+const CompletedJobsAccordion: React.FC<CompletedJobsAccordionProps> = ({
+  completedJobActions,
+}) => {
+  const [isExpanded, toggleAccordion] = useState(false);
```
Khristophor: While I like the idea of trying to be explicit with the naming of what the state update function does, the React convention is to name the state variable and function the same:
```js
const [isExpanded, setIsExpanded] = useState(false);
```
https://reactjs.org/docs/hooks-state.html#tip-what-do-square-brackets-mean

Makes it easier to keep track of which functions are updating what state when there are many items.

resources/assets/js/components/UnclaimedJobCard.tsx
```diff
@@ -4,26 +4,30 @@ import { JobStatus } from "../models/lookupConstants";
 
 export interface UnclaimedJobCardProps {
   title: string;
+  url: string;
```
Khristophor: I added a `Link` type in the JobCard PR that should probably be broken out somewhere else (models/types maybe? @tristan-orourke) that we should use everywhere for consistency. We often seem to be missing things like titles in our anchors.

tristan-orourke: Good point that we're missing titles in anchors, but I'm not sure we need to pass all that in through component props. Wouldn't it make sense to define the anchor title and text in a <FormattedMessage> right in the job card?

Khristophor: Yep, but either way it would be nice to `import { Link } from '../components/Link'` among the various places where we implement an anchor tag so we're not forgetting things.

tristan-orourke: A Link component would be so similar to an anchor element, I'm not convinced its worth it. We added a title attribute in UnclaimedJobCard, its just an empty string, the component wouldn't prevent that.
I feel like what we really want is a linting rule. I did some searching, but couldn't find any for title attributes in anchor tags. We're already using the common accessibility eslint plugin. Maybe we should write our own rule?

Khristophor: I don't think anything will prevent us from deliberately passing empty strings to props, but some sort of component would at least throw a TS error if a required prop (title) is missing. If you can find/write a linting rule for that I'd be onboard.

tristan-orourke: I followed your suggestion @Khristophor and extracted Link into an app-wide interface, and used it in UnclaimedJobCard. Its still easy enough to set the anchor title attribute to an empty string, which we might do here anyway, but we'll have to be deliberate about it.

PR #2311, opened 2019-12-10
Task - laravel breadcrumbs

yonikid15:
For #2276 

### Notes:
- Lightweight breadcrumbs solution 
- Takes the url and builds a list with each segment (eg. `/en/demo/builder-01 = { 'en': 'en', 'demo': 'en/demo'...}`).
- Then replace the key in the list with it's corresponding value in lang file.
- Demo can be viewed in https://talent.test/en/demo/builder-01 page.

UPDATE: I replaced the breadcrumbs on the FAQ page with this dynamic one. https://talent.test/en/faq


resources/views/common/breadcrumbs.html.twig
```diff
@@ -0,0 +1,19 @@
+
+<nav aria-label="breadcrumb" data-c-breadcrumbs data-c-colour="white" data-c-margin="bottom(normal)">
+    <ol>
+      {% for key, url in breadcrumbs %}
+        {% if loop.first %}
+        <li>
+          <a href={{ url(url) }}>{{ breadcrumbs_lang.home }}</a>
```
Khristophor: Missing quotes here on the `href`, and we'll need titles on these anchors too.

yonikid15:
> Left align (like: https://talent.test/en/faq).
> 
> URL probably won't be enough for link text on secondary/tertiary crumbs (i.e. `builder-01`), though might not require a change now.
> 
> Implementing this throughout the site?
> `/faq`, `/settings` ?

There is a lang file for the breadcrumbs where we can customize the text for the breadcrumbs if need be. 

The left align will be handled by the parent template as to make it more modular.

When this gets merged we can add it in wherever needed.

PR #2335, opened 2019-12-11
Redux HR Advisor & Claim Job

### LINKED ISSUE #2176: Task - HR Dashboard - create Redux code for getting, storing and updating HrAdvisor / Claimed Jobs
# Description
We need to store HrAdvisor in Redux store, and we need actions to Get the advisor object, and claim/unclaim jobs via ajax calls.

tristan-orourke:
Resolves #2176 .

### Notes:
- I couldn't figure out how to elegantly connect the idea of a "current user" to the list of HrAdvisors given the existing redux store, so instead I created an api endpoint the claims a job for a specified advisor. That way, the redux code doesn't need to care whether its the current user or not.
- I wrote frontend tests for this one! In hindsight, _some_ of them are useful.
- I would skip redux action tests (hrAdvisorAction.test.ts) next time, since most of what its trying to test is either library functionality (redux-api-middleware) or replaced by mocks anyway.
- The selector and reducer tests I wrote did catch some errors, so 👍


resources/assets/js/fakeData/fakeHrAdvisor.ts
```diff
@@ -0,0 +1,11 @@
+import { HrAdvisor } from "../models/types";
+
+export const fakeHrAdvisor = (overrides: Partial<HrAdvisor> = {}): HrAdvisor => ({
```
Khristophor: `Partial` is pretty neat

tristan-orourke: Yeah!

PR #2339, opened 2019-12-12
Task - Comment model

### LINKED ISSUE #2238: Task - Create Comment model for Job Posters
# Description
We need to create a Comment model associated with Job Posters. It should contain the following fields:
   job_poster_id
   user_id
   created_at (timestamp)
   comment (text)
   location (text)
   context (json)

The location field can be used store a string representing where the comment was added, and the context field can be used to add any extra data we need for a particular comment.

# Acceptance criteria
- [ ] migration
- [ ] Comment model
- [ ] Comments relationship added to JobPoster model.

yonikid15:
Resolves #2238 

### Notes:
https://github.com/GCTC-NTGC/TalentCloud/issues/2238#issuecomment-564601379
- I also created a `CommentTypes` lookup model based off this info from Tristan.


tests/Unit/CommentTest.php
```diff
@@ -0,0 +1,41 @@
...
+     *
+     * @return void
+     */
+    public function testCommentBelongsToAUser()
+    {
+        $user = factory(User::class)->create();
+        $comment = factory(Comment::class)->make(['job_poster_id' => $user->id]);
+
+        $this->assertInstanceOf(User::class, $comment->user);
+    }
```
tristan-orourke: @yonikid15 I don't think we should bother writing tests like these. Eloquent model relationships are a key part of the Laravel framework, we kind of need to trust that they work. Writing and running these tests might be a misuse of our time.

yonikid15: Okay will keep this in mind for next time. I was following this tutorial https://dev.to/tonyfrenzy/part-2-testing-model-relationships-in-laravel-basic-3a44 . I think the goal was to test that the relationships exists and set correctly, so if someone makes any changes it to the relationships it will break.

tristan-orourke: That's fair. Still, I wouldn't bother next time.

PR #2397, opened 2019-12-18
Adds Gtags to job poster items.

joshdrink:
@gggrant worth mentioning the changes to the docker nginx file, because they will also have to be made on the server. Thanks!


gggrant:
Make sure to update the content security policy before running deploy script (to self).

PR #2499, opened 2020-01-03
Finalize Job Index Page

### LINKED ISSUE #2340: Task - HR - Finalize Functionality of Job Index page
# Description
Put together the pieces of the Job Index page for HR advisers and make sure they work. This includes communicating with server.

tristan-orourke:
Resolves #2340.

### Notes:
- I tried out Redux's Hooks api! I think it makes a lot of sense for a root component like this, that just has to manage loading/filtering all the data. Apparently it can't cache child components the same way the `connect` HOC can, so it wouldn't be useful in other cases (eg our current Assessment Plan builder, which uses redux extensively to avoid re-rendering child components when parents are re-rendered).
- I made `localizeField` helper function. I've tried doing something like this before, but I finally understand how to use Typescript to make this function with useful typechecking. I know we have the ongoing translation package refactor. Is it too late to be useful?
- I also made a `readableDateTime` helper function. It just leverages dayjs, but it keeps the logic all in one place. We probably want date- or time-exclusive functions later. Localization included!
- There's not much for loading indicators yet. We just show the null states while jobs are loading.


resources/assets/js/helpers/routes.ts
```diff
@@ -75,3 +75,19 @@ export function managerFaq(locale: string, faqSection?: FaqSection): string {
   }
   return base;
 }
+
+export function hrJobIndex(locale: string): string {
+  return `/${locale}/hr-adviser/jobs`;
```
Khristophor: Should do a global find-replace `adviser` > `advisor` just to ensure consistency

resources/assets/js/models/jobUtil.ts
```diff
@@ -99,3 +102,22 @@ export const getSkillLevelName = (
   }
   return skillLevelName(skill_level_id, skill_type_id);
 };
+
+// TODO: allow for Complete status.
+export const jobStatus = (job: Job): JobStatus => {
```
Khristophor: Shouldn't the server be deciding this?

resources/assets/js/helpers/dates.ts
```diff
@@ -0,0 +1,10 @@
+import dayjs from "dayjs";
+import LocalizedFormat from 'dayjs/plugin/localizedFormat';
+import "dayjs/locale/fr";
+import { Locales } from "./localize";
+
+dayjs.extend(LocalizedFormat)
+
+export const readableDateTime = (locale: Locales, date: Date): string => {
```
Khristophor: How are we handling timezones with this implementation? Everything from the server would be UTC, correct?

Khristophor:
@tristan-orourke only other thing I noticed is we're missing a `relinquish` button on the claimed job card, not sure if that's for another story or not.

resources/assets/js/helpers/routes.ts
tristan-orourke: Ah, you're right, or the other way around. I'll ask about a preferred spelling.

resources/assets/js/models/jobUtil.ts
tristan-orourke: Yeah, you're right. We'll be changing the status system more in a future sprint, but for now I could just add an accessor to the JobPoster model that returns this status id using the same logic. Do you think that's okay for now?

resources/assets/js/helpers/dates.ts
tristan-orourke: Good point, timezones are still ignored here.

resources/assets/js/models/jobUtil.ts
Khristophor: If we aren't yet keeping the `job_poster_status` table in sync with the `JobPoster` model's `status()` method, then yes, we should add something with this logic on the PHP side to be consumed by Javascript.

resources/assets/js/helpers/dates.ts
Khristophor: We should be able to [determine the timezone on the fly](https://stackoverflow.com/a/34602679) (if we want to display based on browser/system info), or we can display an explicit timezone like we're doing on the Twig side.

tristan-orourke:
> @tristan-orourke only other thing I noticed is we're missing a `relinquish` button on the claimed job card, not sure if that's for another story or not.

It doesn't show up on @joshdrink 's [mock-up](https://joshbeveridge.ca/tc-manager-dash/hr-adviser/), I think it was decided to only put that button on the summary page.

joshdrink:
> 
> 
> > @tristan-orourke only other thing I noticed is we're missing a `relinquish` button on the claimed job card, not sure if that's for another story or not.
> 
> It doesn't show up on @joshdrink 's [mock-up](https://joshbeveridge.ca/tc-manager-dash/hr-adviser/), I think it was decided to only put that button on the summary page.

I indeed believe this is true :)

resources/assets/js/helpers/dates.ts
tristan-orourke: Whoah! dayjs seems to be converting it from utc to my browsers local timezone, on its own!

tristan-orourke: Could timezones actually be easy, for once??

Khristophor: The question is, do we want that? In most other places we're displaying EST/EDT explicitly haha.

tristan-orourke: > The question is, do we want that?
😭

tristan-orourke: I'll try to get the right EST/etc code to display explicitly.

tristan-orourke: Sounds like Gray thinks local time is fine.

PR #2514, opened 2020-01-06
Bugfix/disable department

### LINKED ISSUE #2341: Bug - Manager/HR: Being Able to Change Department
# Description

We need to make sure that the ability to change your department has been removed from the Manager and HR facing UI. The only way these users should be able to change their department is by contacting Talent Cloud directly.

We should do a review of:
- the manager profile
- the job poster builder (step 1?)

We should ensure that the flow for managers found in issue #1274 applies to HR (e.g. if the user signs up as an applicant and THEN visits the HR portal, we need to ensure we ask for their gov info in the same manner).

yonikid15:
Resolves #2341 

### Notes:
- Remove department select box:
  - Manager Profile
  - Welcome page of Job Poster Builder
- Added message with anchor to email talent cloud.   

### Update:
- [x] Remove select box and just use `<p>` tag
- [x] Instead link to https://talent.test/en/settings#government in message


tests/Browser/2-critical.testcafe.js
```diff
@@ -42,7 +42,7 @@ test("Applicant Profile - My Skills", async t => {
     .expect(Selector("h3").withText("Writing my application").visible)
     .ok()
     .click(Selector("button").withText("Got it!"))
-    .wait(300)
+    .wait(1234)
```
Khristophor: I think this is covered by #2516

resources/assets/js/components/JobBuilder/Intro/IntroForm.tsx
```diff
@@ -343,6 +361,7 @@ const IntroForm: React.FunctionComponent<IntroFormProps &
                     grid="base(1of1)"
                     component={SelectInput}
                     required
+                    disabled
```
Khristophor: Should we just replace this with a `<p>` or similar? Are we looking to make it editable for different roles in the future?

tests/Browser/2-critical.testcafe.js
yonikid15: Ya so I fixed the broken test here first. Then noticed it in #2499 PR as well. Thats why I made a PR to fix it.

resources/assets/js/components/JobBuilder/Intro/IntroForm.tsx
yonikid15: Yea we definitely could. @joshdrink what do you think? I wasn't sure what the final product should look like.

yonikid15: Just talked to Josh. Next steps: 
- Remove select box and just use `<p>` tag
- Instead link to https://talent.test/en/settings#government in message

yonikid15:
@Khristophor I went with just checking if the department element exists or not.

PR #2517, opened 2020-01-06
Task - Allow frontend to request a list of jobs, with filters

### LINKED ISSUE #2500: Task - Allow frontend to request all jobs in a department from API
# Description
We have an api endpoint (and redux store) for getting individual jobs, but no way to request a list of all jobs, or all jobs in a department.
We must create a new API endpoint and redux code for getting a list of jobs, preferably with the option of filtering by department.

tristan-orourke:
Resolves #2500.


resources/assets/js/api/base.ts
```diff
@@ -1,5 +1,6 @@
 import dayjs from "dayjs";
 import rootAxios from "axios";
+import { string } from "yup";
```
Khristophor: Should this be here? Or are you just using the generic string type for typescript?

resources/assets/js/components/HRPortal/JobIndexHr.tsx
```diff
@@ -94,13 +94,13 @@ const CompletedJobsAccordion: React.FC<CompletedJobsAccordionProps> = ({
   );
 };
 
-interface JobIndexPageProps {
+interface JobIndexHrProps {
```
Khristophor: Is there going to be a functional difference between the HR Portal index and any other index? I'd also like to come up with some sort of naming convention to follow for these types of files.

On the JPB we have the parent folder containing the sections, and if each section has a need for multiple files we subfolder those... the "container" component have the `SectionPage` naming convention in those sections, so I'd like to either follow that or figure out something else.

resources/assets/js/api/base.ts
tristan-orourke: Oops! Nah, just something my IDE added by mistake.

resources/assets/js/components/HRPortal/JobIndexHr.tsx
tristan-orourke: The Job index on all three portals is distinct. Only HR can claim jobs, Managers can edit them, and applicants can only view the final poster.

I've been following the `___Page` for containers patterns. I don't see JobIndexHr as the container component, I've created a JobIndexHrPage component in the other branch I'm working on.

PR #2522, opened 2020-01-07
Bugfix - misc bugfixes

yonikid15:
### Notes:
- Modal was overflowing outside screen. Might be caused by a clone update.
- ScrollToTop needed children prop in useEffect dependencies.
- Recursively update wordcounter if new skill button is triggered.

### DIFF
--- a/resources/assets/js/components/ApplicantSkills/SkillsWordCounter.tsx
+++ b/resources/assets/js/components/ApplicantSkills/SkillsWordCounter.tsx
@@ -96,11 +96,16 @@ const addHardSkillButton: HTMLElement | null = document.getElementById(
 );
 
 const updateWordCounters = (): void => {
+  if (addSoftSkillButton) {
+    addSoftSkillButton.addEventListener("click", updateWordCounters);
+  }
+
+  if (addHardSkillButton) {
+    addHardSkillButton.addEventListener("click", updateWordCounters);
+  }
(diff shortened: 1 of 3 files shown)


resources/assets/js/components/ApplicantSkills/SkillsWordCounter.tsx
```diff
@@ -96,12 +96,17 @@ const addHardSkillButton: HTMLElement | null = document.getElementById(
 );
 
 const updateWordCounters = (): void => {
+  if (addSoftSkillButton) {
+    addSoftSkillButton.addEventListener("click", updateWordCounters);
```
Khristophor: Do we have anything we can hook into to remove these event listeners?

yonikid15: The add skill buttons are on the laravel side so I'm not sure how other than this.

tristan-orourke: Yeah, in this case I'm not sure how else to do it.

Khristophor: Alrighty, my only concern here is having a bunch of event listeners hanging around in the browser after heavy use, but we can keep an eye on it in the future.

PR #2592, opened 2020-01-14
Set up CI with Azure Pipelines

vid:


gggrant:
Let's open a PR when it's closer to ready.

PR #2595, opened 2020-01-14
Set up correct HR Advisor auth policies.

tristan-orourke:
### Notes:
- HR Advisors can view and claim jobs if they are in the same department and the job has been sent for review.
- HR advisors can view the job summary, view applicants, comments, and applications, if they still have permission to view the job and have already claimed it.
- The job must be closed for advisors to view applications and applicants


app/Policies/JobPolicy.php
```diff
@@ -113,25 +111,44 @@ public function submitForReview(User $user, JobPoster $jobPoster)
...
+     * @param JobPoster $jobPoster
+     * @return void
+     */
+    public function manage(User $user, JobPoster $jobPoster)
+    {
+        return ($user->isManager() &&
+            $jobPoster->manager->user->id == $user->id) ||
+            ($user->isHrAdvisor()
+                && $this->view($user, $jobPoster)
+                && $user->hr_advisor->claimed_job_ids->includes($jobPoster->id));
```
yonikid15: I got an error for this. I don't know `includes()` exists anymore. I switched it to `in_array()`.

tristan-orourke: Ah, you're right. I'm trying to use laravel collection methods - it should be [contains](https://laravel.com/docs/5.2/collections#method-contains).

tristan-orourke:
There are actually failing tests that need to be fixed here.

yonikid15:
> There are actually failing tests that need to be fixed here.

Ya I assumed it was the langFiles test, but saw there were more so I'm doing a re-review

tristan-orourke:
@yonikid15 There were a couple things wrong with claimed_job_ids and the unclaim route, I've fixed those.

PR #2674, opened 2020-01-21
Task - Applicant Review page activity feed

### LINKED ISSUE #2321: Task - Manager/HR - Add comment components to list of applicants page
# Description
Managers and HR need the ability to leave comments on various sections of the Applicant Review page. 

**Please select the users this UI will apply to.**
- [ ] Administrators
- [ ] [Applicants](https://github.com/GCTC-NTGC/TalentCloud/issues/1548)
- [x] [Hiring Managers](https://github.com/GCTC-NTGC/TalentCloud/issues/1549)
- [x] [HR Advisers](https://github.com/GCTC-NTGC/TalentCloud/issues/1550)

## Assumptions
- HRs and Managers will have the option to choose which section of the Review Applicants page they want to comment on using a pre-populated drop-down list.
- HRs and Managers will be required to Leave a comment.
- HRs will have the option to Select a comment type whereas Managers do not have this option.
- Once comments are made, users will have the ability to show/hide comments making it more usable. 

**Select a section options include:**
- Generic
- Under Consideration
- Optional Consideration
- No Longer Under Consideration

**Select a comment type options include:**
- Question
- Recommended Action 
- Required Action

## Version 1
[XD Wireframe](https://xd.adobe.com/view/82df8c10-237f-4158-7a93-bbde66641386-f4ee/rl)

yonikid15:
Resolves #2321 

### Notes:
- I renamed the ActivityFeed function to ActivityList since it didn't really make sense, and I made a new component called ActivityFeed that has the CommentForm and AcitivtyList. 
- Also includes refactoring of localized constants to remove duplicates.


resources/assets/js/translations/locales/fr.json
```diff
@@ -2,18 +2,28 @@
...
-  "activityfeed.locations.review.impact": "Constructeur d'Affiches > Impact",
-  "activityfeed.locations.review.langRequirements": "Constructeur d'Affiche > Exigences linguistiques",
-  "activityfeed.locations.review.skills": "Constructeur d'Affiches > Compétences",
-  "activityfeed.locations.review.tasks": "Constructeur d'Affiches > Taches",
-  "activityfeed.locations.summary": "Résumé de l'emploi",
+  "activityfeed.loadingIconText": "Nombre d'activités est en cours de chargement...",
+  "activityfeed.locations.applicantReview.generic": "Général",
+  "activityfeed.locations.applicantReview.notUnderConsideration": "No Longer Under Consideration",
+  "activityfeed.locations.applicantReview.optionalConsideration": "Optional Consideration",
+  "activityfeed.locations.applicantReview.underConsideration": "Under Consideration",
```
Khristophor: Should get these localized

yonikid15: I'm don't think we have translations for these right now, but I yes they should be. I'll check with Gray if the deepl.com translations will suffice.

Khristophor: Not having translations led me to #2882 which I think has come up a few times. Might want to tackle that first so we can get the related text translated as well.

yonikid15: Ya forsure, I can do that. Also, looks like there might be a change to the English copy, but will find out tomorrow.

resources/assets/js/components/ActivityFeed.tsx
```diff
@@ -1,182 +1,133 @@
...
 
 interface ActivityFeedProps {
   jobId: number;
   isHrAdvisor: boolean;
-  comments: Comment[];
-  handleFetchComments: (jobId: number) => Promise<void>;
-  filterComments?: (comment: Comment) => boolean;
+  generalLocation: string;
+  locationMessages: { [LocationId: string]: MessageDescriptor };
+  totalActivities: number;
```
tristan-orourke: I think you should change the interface slightly. ActivityFeed should accept a list of comments, count them inside the component instead of recieving the count as a prop, and pass the comments to ActivityList. The exported, redux-connected component should take an optional filterComments prop, just like ActivityList.

a) This would make it impossible for this component to be given a count which doesn't accurately reflect what's shown in ActivityList.
b) If other components are including ActivityFeed instead of ActivityList, they need a way to filter the comments.

resources/assets/js/components/HRPortal/JobReviewHrPage.tsx
```diff
@@ -73,23 +75,30 @@ const JobReviewHrPage: React.FunctionComponent<JobReviewHrPageProps> = ({
...
               defaultMessage="Just a heads up! We've rearranged some of your information to help you understand how an applicant will see it once published."
               description="Description under primary title of review section"
             />
           </p>
-          <JobReviewActivityFeed jobId={job.id} isHrAdvisor />
+          <ActivityFeed
+            jobId={job.id}
+            isHrAdvisor
+            generalLocation={LocationId.jobGeneric}
+            locationMessages={jobReviewLocations}
```
tristan-orourke: Once ActivityFeed accepts a filterComments function, we should get this to filter out comments not associated with the job builder (eg Application comments).

resources/assets/js/components/JobBuilder/Review/JobReview.tsx
```diff
@@ -784,15 +786,20 @@ export const JobReview: React.FunctionComponent<JobReviewProps &
...
             understand how an applicant will see it once published."
             description="Description under primary title of review section"
           />
         </p>
-        <JobReviewActivityFeed jobId={job.id} isHrAdvisor={false} />
+        <ActivityFeed
+          jobId={job.id}
+          isHrAdvisor={false}
+          generalLocation={LocationId.jobGeneric}
+          locationMessages={jobReviewLocations}
```
tristan-orourke: Once ActivityFeed accepts a filterComments function, we should get this to filter out comments not associated with the job builder (eg Application comments).

resources/assets/js/components/ApplicationReview/ReviewApplications.tsx
```diff
@@ -131,6 +136,16 @@ const ReviewApplications: React.StatelessComponent<ReviewApplicationsProps> = ({
           />
         </div>
       </div>
+      <div data-clone>
+        <div data-c-margin="bottom(1)">
+          <ActivityFeed
```
tristan-orourke: Once ActivityFeed accepts a filterComments function, this should only allow comments with the four Applicant locationIds.

resources/assets/js/components/ActivityFeed.tsx
yonikid15: Okay will do. For a) though, the count would should always be accurate since it's being grabbed from the root state.

PR #2705, opened 2020-01-22
Bugfix - Change preview job permission 

### LINKED ISSUE #2704: Bug - HR Portal - Clicking on a job
# Description

Clicking on a job card on the Jobs page takes me to a 403 error. I don't think the job card should be clickable, users should only be able to click on the Claim This Job button. 

# Technical Info

**Device:** Surface Tablet
**Operating System:** Windows
**Browser:** Chrome

# Steps to Reproduce
1. Go to the HR portal and login
2. Go to the Jobs Page
3. Click on any job in the All Jobs section

# Images
Capture

Capture2

yonikid15:
Resolves #2704 

### Notes:
- This is a happening since the HR Advisors can only view jobs that they have claimed. So we removed the link from the Unclaimed job cards. Now, the advisor can only view the job once they have claimed it.

UPDATE: Instead I changed the permissions to open the Job Preview page for hr advisors.


resources/assets/js/components/UnclaimedJobCard.tsx
```diff
@@ -28,89 +27,79 @@ const UnclaimedJobCard: React.FunctionComponent<UnclaimedJobCardProps> = ({
...
+              defaultMessage="Hiring Managers: "
+            />
+
+            {hiringManagers.map((manager, index): string => {
+              const comma =
+                hiringManagers.length !== 1 &&
+                index + 1 !== hiringManagers.length
+                  ? ","
+                  : " ";
+              return `${manager}${comma} `;
```
tristan-orourke: This could be replaced by `{hiringManagers.join(", ")}`.

resources/assets/js/components/UnclaimedJobCard.tsx
```diff
@@ -1,11 +1,10 @@
...
 import { JobStatus } from "../models/lookupConstants";
-import { Link } from "../models/app";
 import { jobStatus } from "../models/localizedConstants";
 
 export interface UnclaimedJobCardProps {
-  jobLink: Link;
+  jobTitle: string;
   createdAt: string;
   status: JobStatus;
   hiringManagers: string[];
```
tristan-orourke: Its only possible for a job to have a single manager at a time.

resources/assets/js/components/UnclaimedJobCard.tsx
```diff
@@ -28,89 +27,79 @@ const UnclaimedJobCard: React.FunctionComponent<UnclaimedJobCardProps> = ({
...
-              {hiringManagers.map((manager, index): string => {
+              {hrAdvisors.map((advisor, index): string => {
                 const comma =
-                  hiringManagers.length !== 1 &&
-                  index + 1 !== hiringManagers.length
+                  hrAdvisors.length !== 1 && index + 1 !== hrAdvisors.length
                     ? ","
                     : " ";
-                return `${manager}${comma} `;
+                return `${advisor}${comma} `;
```
tristan-orourke: As above, use join instead of map.

yonikid15:
> Alternatively, instead of removing the link we could change the permission required to preview jobs. I think we should use the "view" permission instead of the "manage" permission for `hr_advisor.jobs.preview` route. HR advisors already have this permission for all then Unclaimed Job Card jobs.

That's actually a better idea. That way they're not claiming and relinquishing jobs just to view them.

resources/assets/js/components/UnclaimedJobCard.tsx
yonikid15: In the wireframe there were multiple. However, if that's the case then I can remove all of the `hiringManagers.map()` stuff.

gobyrne:
Does this mean HR advisors will be able to view unpublished (but shared with HR) job posters without first claiming it? @yonikid15 @tristan-orourke 

If so, it's acceptable, since there's no personal information there, but maybe not what we want long-term.

yonikid15:
> Does this mean HR advisors will be able to view unpublished (but shared with HR) job posters without first claiming it? @yonikid15 @tristan-orourke
> 
> If so, it's acceptable, since there's no personal information there, but maybe not what we want long-term.

Yes, so once a manager creates a job in JPB. It will be viewable by HR, without needing to claim it first.

app/Http/Controllers/JobController.php
```diff
@@ -165,11 +165,19 @@ function ($value, $key) {
                 'text' => $jobLang['apply']['edit_link_label'],
             ];
         } elseif (WhichPortal::isHrPortal()) {
-            $applyButton = [
-                'href' => route('hr_advisor.jobs.summary', $jobPoster->id),
-                'title' => null,
-                'text' => Lang::get('hr_advisor/job_summary.summary_title'),
-            ];
+            if ($jobPoster->hr_advisors->get($user->id)) {
```
tristan-orourke: I'm not sure [get](https://laravel.com/docs/5.8/collections#method-get) is the correct method here. This will only work properly if the collection is keyed by object id, which I'm not sure is the case. I'd also recommend a `!== null` for explicitness. 

But instead, I'd use [contains](https://laravel.com/docs/5.8/collections#method-contains) with a key value pair, ie `$jobPoster->hr_advisors->contains('id', $user->id)`

resources/assets/js/components/UnclaimedJobCard.tsx
```diff
@@ -92,13 +84,7 @@ const UnclaimedJobCard: React.FunctionComponent<UnclaimedJobCardProps> = ({
...
                   defaultMessage="HR Advisors: "
                 />
-                {hrAdvisors.map((advisor, index): string => {
-                  const comma =
-                    hrAdvisors.length !== 1 && index + 1 !== hrAdvisors.length
-                      ? ","
-                      : " ";
-                  return `${advisor}${comma} `;
-                })}
+                {hrAdvisors.join(",")}
```
tristan-orourke: You probably want to add a space after the comma, `.join(", ")`

yonikid15: Yup :p

PR #2817, opened 2020-02-03
Feature - hidden job option

### LINKED ISSUE #2766: Feature - Applicant/Admin - Hidden Job option
# Description

Option for admins that lets them publish a job in a way that it does not show up on the browse jobs page. It still can be viewed by anyone with the direct link.

# Images
Place image attachments (wireframes, etc.) here.

- [ ] This task is complex and needs acceptance criteria.

# Acceptance criteria
- [ ] Requirement one
- [ ] Requirement two
- [ ] Requirement three

yonikid15:
Resolves #2766 

### Notes:
- Allows admin to "soft publish" a job. 
- This will make the job poster viewable via direct link.


app/Http/Controllers/Admin/JobPosterCrudController.php
```diff
@@ -67,6 +67,7 @@ public function setupListOperation()
             'type' => 'model_function',
             'function_name' => 'status'
         ]);
+
         $this->crud->addColumn([
```
Khristophor: Should probably add another column to see which Jobs are "soft published" from the index view

PR #2832, opened 2020-02-04
Refactor: extract frontend localization to function

tristan-orourke:
This adds 3 helper functions:
- Two are for getting a field from a model (eg `city` from a `job`) in the right locale, depending on whether the result may be null or not.
- One function to replace the "get locale from react-intl, throw exception if not en or fr" pattern we have all over the place.

I've tried to replace every instance of getting a localized field with the use of a function (except for cases where we're constructing a new model or using both locales). 
I've only replaced getting locale from react-intl in a few places so far.

### Notes:
- Having all localization go through a couple functions will make it easy to do things like add null checking for everything.
- The type checking on these functions is actually helpful. You should get code-completion options, for only the localizable fields.
- This assumes that all the models use the same pattern for localization, but it also makes it easy to change (eg if we had to do something like the common-language-package refactor again).


tristan-orourke:
Recreated this PR with a cleaner git history #2833

PR #2833, opened 2020-02-04
Refactor: Always use function to extract localized values from ts objects

tristan-orourke:
This adds 3 helper functions:
- Two are for getting a field from a model (eg `city` from a `job`) in the right locale, depending on whether the result may be null or not.
- One function to replace the "get locale from react-intl, throw exception if not en or fr" pattern we have all over the place.

I've tried to replace every instance of getting a localized field with the use of a function (except for cases where we're constructing a new model or using both locales). 
I've only replaced getting locale from react-intl in a few places so far.

### Notes:
- Having all localization go through a couple functions will make it easy to do things like add null checking for everything.
- The type checking on these functions is actually helpful. You should get code-completion options, for only the localizable fields.
- This assumes that all the models use the same pattern for localization, but it also makes it easy to change (eg if we had to do something like the common-language-package refactor again).

### DIFF
--- a/resources/assets/js/components/AssessmentPlan/RatingGuideClipboard.test.ts
+++ b/resources/assets/js/components/AssessmentPlan/RatingGuideClipboard.test.ts
@@ -19,7 +19,7 @@ import {
   criteriaType,
 } from "../../models/localizedConstants";
 import { clipboardData, ClipboardTableRowProps } from "./RatingGuideClipboard";
-import { localizeField } from "../../helpers/localize";
+import { localizeFieldNonNull } from "../../helpers/localize";
 
 const jediSkill: Skill = fakeSkill();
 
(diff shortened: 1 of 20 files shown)


resources/assets/js/helpers/localize.ts
```diff
@@ -1,15 +1,27 @@
...
+): string | null {
+  const a = model[field];
+  return model[field][locale];
+}
+export function localizeFieldNonNull<T>(
+  locale: Locales,
+  model: T,
+  field: TranslatableKeysNonNull<T>,
+): string {
+  const a = model[field];
```
Khristophor: I think we'll want to do something with these `const a = model[field];`'s, either returning `a[locale]` or removing the `const` declaration.

resources/assets/js/components/AssessmentPlan/RatingGuideClipboard.test.ts
```diff
@@ -19,6 +19,7 @@ import {
   criteriaType,
 } from "../../models/localizedConstants";
 import { clipboardData, ClipboardTableRowProps } from "./RatingGuideClipboard";
+import { localizeField } from "../../helpers/localize";
```
yonikid15: Looks like all the methods in this file need to be changed to `localizeFieldNonNull()`

PR #2845, opened 2020-02-04
Bugfix - Move all "screened in" applicants to Under Consideration category

### LINKED ISSUE #2345: Bug - After the manager has "screened-in" a candidate that was originally under "optional consideration" should be moved into "under consideration"
# Description
Managers have mentioned that when a "screen-in" candidates remain under "optional consideration", they could miss them for further assessment.

More information to follow...

# Steps to Reproduce

# Images
Place image attachments here.

yonikid15:
Resolves #2345


resources/assets/js/components/ApplicationReview/helpers.ts
```diff
@@ -15,6 +15,16 @@ export function isScreenedOut(application: Application): boolean {
...
 
+/**
+ * Returns true if application has been screened in.
+ */
+export function isStillIn(application: Application): boolean {
+  return application.application_review &&
+    application.application_review.review_status
+    ? application.application_review.review_status.name === "still_in"
+    : false; // non-reviewed applications have not been screened-out yet
+}
```
tristan-orourke: Check this out! [Typescript optional chaining.](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#optional-chaining)

You can replace all that with `application?.application_review?.review_status?.name === "still_in"`. If any of the steps is null or undefined, it will cut evaluation short and return `undefined`. Which will, of course not, equal "still_in".

yonikid15: Awesome thanks!

PR #2878, opened 2020-02-06
Improved logic for holiday message

gggrant:
When I pointed this out, nobody cared.


Khristophor:
Thank you for using a boolean and not trying to rely on dates

PR #2894, opened 2020-02-07
Task - sorting unclaimed jobs 

### LINKED ISSUE #2718: Task - HR - Reverse order of unclaimed Job Cards on /hr/jobs page
# Description
Reverse the order of unclaimed Job Cards on /hr/jobs page so that the most recently shared ones are at (or closer to) the top. Ideally this would be ordered by [dateSharedWithHR] but simply in reverse order that the job posters were created is ok.

yonikid15:
Resolves #2718 

### Notes:
- Order the unclaimed jobs by date it is shared with HR (review_requested_at).
- Changed the heading on job card from "Created" to "Received".


resources/assets/js/components/UnclaimedJobCard.tsx
```diff
@@ -48,11 +48,17 @@ const UnclaimedJobCard: React.FunctionComponent<UnclaimedJobCardProps> = ({
...
                     description="Header for when Job Poster was created."
-                    defaultMessage="Created: "
+                    defaultMessage="Received: "
                   />
-                  {createdAt}
+                  {reviewRequested || (
+                    <FormattedMessage
+                      id="openJobCard.error"
+                      description="Error getting time review was requested."
+                      defaultMessage="Error"
```
Khristophor: I feel like we should have something more helpful than `Received: Error` displayed if the Job doesn't have a `review_requested_at` field available. Probably an ask for a designer/UXer.

resources/assets/js/components/HRPortal/__snapshots__/JobIndexHr.test.tsx.snap
```diff
@@ -216,10 +215,10 @@ exports[`Job Index Page should render JobIndexHr correctly 1`] = `
             }
           }
           key="4"
+          reviewRequested="2019-MAY-02"
```
Khristophor: Is this snapshot how the fixtures are rendered? It doesn't look like they're in newest -> oldest by review here.

yonikid15: Ahh yes, the unclaimed jobs list gets sorted in `JobIndexHrPage.tsx` which is a wrapper before passing data to `JobIndexHr.tsx`. I can change it to use JobIndexHrPage component instead.

yonikid15: Actually it's easier if I just sort it within the after it gets passed into `JobIndexHr.tsx`

PR #2908, opened 2020-02-11
Add User name and role to Activity Feed

Khristophor:


Khristophor:
> The comment feed errors until you add save names to the user roles. I like having the Admin portal available to change them, but it might be safest to initialize them with a migration.
> 
> To merge ActivityFeed, you probably want to more your code into the new ActivityList component.

Working on the merge now, I'll add a migration for the role names.

Khristophor:
@tristan-orourke should be good to go now

PR #2936, opened 2020-02-14
Overhaul Job Poster Status and status transitions

### LINKED ISSUE #2189: Task - Job Poster - Add new 'status' column to table. Update value accordingly during poster's life-cycle.
# Description
- Add in a new column called `status` to the `job_posters` table. This column will hold the corresponding status of the Job Poster over it's life cycle. The possible values include:

  - Draft | default value
  - In Review | TBD
  - In Translation | TBD
  - In Approval | TBD
  - Approved | TBD
  - Ready to Post | TBD
  - Poster Live | TBD
  - Assessment Phase | TBD
  - Process Complete | TBD

EDIT: Must reflect all possible states of this chart: https://whimsical.com/VxJf9aJveb3gY7y2Jopep7

- At each point where the status of the job is updated (event triggers, or after duration in time). The correct status should be reflected in the job poster.

- [ ] This task requires refactoring unit tests.

# Acceptance criteria
- [ ] New column named `status` in `job_posters` table.
- [ ] The value matches the corresponding status throughout the Job Poster's life-cycle.

tristan-orourke:
Resolves #2189.
This does quite a lot.
Its based on this graph https://whimsical.com/VxJf9aJveb3gY7y2Jopep7

### Notes:
- Theres now a job_poster_statuses table describing all possible states, and a job_poster_status_transitions table, which records rules about which transitions are possible and by which users.
-  These tables also contain localized text for display.
- Localized text is for statuses is editable in admin portal, and the transitions are fully editable.
- Changing job status is done through its own endpoints. There is an api endpoint, and endpoints for form requests on the hr and manager portals. job_poster_status_transitions is used to validated each transition.
- Every time job status changes, the transition is recorded in job_poster_transition_histories table. This is accomplished via an event listener.
- Buttons have been added along the side of the Job Summary page that trigger currently-legal transitions. They're disabled if the user doesn't have the right permission. We may want to add more description of what they do or why they're disabled.
- Client-side code gets all the possible JobPosterStatuses with redux & an api request, and compares them to job.job_poster_status_id to get localized names/descriptions of the statuses. **Unfortunately, local constants are still used to check for certain statuses (ie draft, or completed). This allows for the possibility of frontend/backend de-synchronization, but I'm not sure how to avoid this.**
- Admins can change the job status through the admin portal, or by manually navigating to a Job Summary page on the HR or Manager portal.


resources/assets/js/models/lookupConstants.ts
```diff
@@ -129,11 +129,15 @@ export const LocationId = {
 
 export enum JobStatus {
   Draft = 1,
-  Review = 2,
-  Approved = 3,
-  Open = 4,
-  Closed = 5,
-  Complete = 6,
+  ReviewManager = 2,
```
Khristophor: As chatted, we could update these to string literal enums:
```ts
ReviewManager = "review_manager"
```

app/Services/JobStatusTransitions.php
```diff
@@ -0,0 +1,139 @@
+<?php
+
+namespace App\Services;
+
+use App\Models\User;
+
+class JobStatusTransitions
+{
+    protected $transition_graph = [
+        'states' => [
```
Khristophor: I think it makes sense to store these states and transitions in the database, then we could make actual relationships between things like owners and user roles, and provide localized text for names and/or descriptions.

I used a `key` and `name` field convention for storing the text key and localized name respectively on the User Role edits I made.

Khristophor:
And I think the other thing we talked about was returning the `key`/`name` for the Job Status through the API so the frontend could use them with the Typescript enums for statuses.

app/Http/Controllers/JobStatusController.php
```diff
@@ -0,0 +1,50 @@
...
+        $transition->from_job_poster_status_id = $fromStatus->id;
+        $transition->to_job_poster_status_id = $toStatus->id;
+        $transition->save();
+
+        return $request->ajax()
+            ? response()->json(['status' => 'ok'])
+            : back();
+    }
+
+    public function setJobStatus(Request $request, JobPoster $jobPoster, string $status)
```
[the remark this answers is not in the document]
tristan-orourke: I separated it because I wanted to allow flexibility in how the entry point gets the destination status. I wasn't sure if the url should specify the status to transition to (the current solution) or whether that should be part of a payload (ie {status: 'completed'}). Do you have any preference?

app/Http/Controllers/JobStatusController.php
```diff
@@ -0,0 +1,50 @@
...
+        // Save transition history.
+        $transition = new JobPosterStatusHistory();
+        $transition->job_poster_id = $job->id;
+        $transition->user_id = $user->id;
+        $transition->from_job_poster_status_id = $fromStatus->id;
+        $transition->to_job_poster_status_id = $toStatus->id;
+        $transition->save();
+
+        return $request->ajax()
+            ? response()->json(['status' => 'ok'])
```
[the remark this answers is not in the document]
tristan-orourke: Yeah. In this case, I realized it should return the JobPosterResource for api requests. It's nice having that ready to go.

resources/assets/js/models/lookupConstants.ts
tristan-orourke: I was starting to think this would be more trouble than its worth, but I think I see how to get this to work now. Will finish this up tomorrow morning.

app/Http/Controllers/JobStatusController.php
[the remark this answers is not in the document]
Khristophor: I think payload probably makes more sense? That way we don't have to maintain `n` routes and changes to the statuses can be more seamless.

resources/assets/js/models/lookupConstants.ts
tristan-orourke: I've pushed this up now.

app/Http/Controllers/JobStatusController.php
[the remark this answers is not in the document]
tristan-orourke: If its only about maintaining routes, the current setup is pretty good:
```
Route::post(
        'jobs/{jobPoster}/set-status/{status}',
        'JobStatusController@setJobStatus'
    )
```

app/Http/Controllers/JobSummaryController.php
```diff
@@ -62,39 +64,49 @@ public function show(Request $request, JobPoster $job)
...
         $portal = '';
-
         if (WhichPortal::isHrPortal()) {
             $portal = 'hr';
         } elseif (WhichPortal::isManagerPortal()) {
             $portal = 'manager';
         }
 
+        $transitionManager = new JobStatusTransitionManager();
+        $transitionToButton = function (JobPosterStatusTransition $transition) use ($user, $job, $transitionManager) {
```
Khristophor: Not suggesting you change this, but this is a very Javascript way of writing PHP, and something Scrutinizer wouldn't be a fan of.

Traditionally you would do something like
```php
foreach ($transitionManager->legalTransitions($job->job_poster_status->key)) {
    array_push($buttonGroups, $this->transitionToButton($transition, $user, $job, $transitionManager));
}
```

With the `transitionToButton` function being defined outside of the `show` method here.

Again, not saying you need to refactor, and I know it's a byproduct of hopping between languages frequently, just something to keep in mind.

tristan-orourke: Thanks. This is less a byproduct of hopping between languages, and more about me being less familiar with PHP conventions in the first place. Most of my reading is about javascript, or even more functional languages.

routes/web.php
```diff
@@ -393,6 +393,13 @@ function (): void {
                             ->where('jobPoster', '[0-9]+')
                             ->middleware('can:delete,jobPoster')
                             ->name('manager.jobs.destroy');
+                        Route::post(
+                            'jobs/{jobPoster}/set-status/{status}',
```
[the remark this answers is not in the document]
tristan-orourke: Good point. Renamed the routes to `jobs/{jobPoster}/status/{status}`. Switched the api route to a PUT, since that seems more appropriate, but left the HR and Manager portal POST routes for form requests. Also made sure that the controller can gracefully handle a request to set the status to what the job already has.

tristan-orourke:
> Is it safe for Admins to have access to create/edit/delete Job Status Transitions? Seems like there's a lot of risk of breaking Job Posters with having all of these fields available.
> Screen Shot 2020-03-02 at 4 39 37 PM

The code should be able to handle whatever transitions we have here, but there's definitely the risk of breaking the business logic we've settled on. Its a compromise: do we make it easy to change the business logic whenever we want, or do we make it hard to change it accidentally?

I can disable the critical fields for now. Leaving it easy to change later if we want.

Khristophor:
> > Is it safe for Admins to have access to create/edit/delete Job Status Transitions? Seems like there's a lot of risk of breaking Job Posters with having all of these fields available.
> > Screen Shot 2020-03-02 at 4 39 37 PM
> 
> The code should be able to handle whatever transitions we have here, but there's definitely the risk of breaking the business logic we've settled on. Its a compromise: do we make it easy to change the business logic whenever we want, or do we make it hard to change it accidentally?
> 
> I can disable the critical fields for now. Leaving it easy to change later if we want.

I think disabling for now makes sense. If we have a need for it in the future we can expose the functionality (with the appropriate warnings :P)

PR #2950, opened 2020-02-18
Manager/HR - Fix gov_email rules on first visit pages

### LINKED ISSUE #2873: Bug - HR / Manager portal - GC email requirement
# Description
When an HR advisor who already has a manager account tries to log-in to the HR advisor account for the first time, they are asked to provide a government email. Because they have already provide that information for the manager account, however, they would receive an error message when trying to enter their email again. We currently work around it but asking the HR advisor to add a "+" to their email address.

gggrant:
Resolves #2873.


resources/views/auth/manager_registration_fields.html.twig
```diff
@@ -6,21 +6,21 @@
...
 		<i class="fa fa-caret-down"></i>
 		<select id="department" name="department" required>
-			<option selected value="" disabled>{{ copy.department_null_selection }}</option>
+			<option {% if user.department_id == null %} selected {% endif %} value="" disabled>{{ copy.department_null_selection }}</option>
 			{% if not_in_gov_option %}
 				<option value="{{ not_in_gov_option.value }}">{{ not_in_gov_option.name }}</option>
 			{% endif %}
 			{% for dept in departments %}
-				{% if old('department') == dept.id %}
+				{% if user.department_id == dept.id %}
```
yonikid15: Curious as to why this got changed? It's supposed to grab the old input if validation fails, so users don't need to re-enter data.

app/Http/Controllers/Auth/FirstVisitController.php
```diff
@@ -41,9 +45,18 @@ public function showFirstVisitManagerForm()
...
+        $data = $request->validate([
+            'department' => 'required|integer',
+            'gov_email' => [
+                'nullable',
+                // gov_email is required unless department is set to 0 (Not in Government)
+                'required_unless:department,0',
+                'email:rfc',
+                'max:191',
+                // gov_email may match existing email for this user, must be unique if changed.
+                Rule::unique('users', 'email')->ignore($request->user()->id)
```
yonikid15: Getting an error when registering and trying to use a `gov_email` that already exists. Pretty sure it's happening because the `user()` trying to register doesn't exist yet. 
image

gggrant: Fixed this. See the rules on RegistrationValidator vs. FirstVisitController.

resources/views/auth/manager_registration_fields.html.twig
gggrant: I was getting unwanted results with that, if the user had already selected a department. Just added or logic and seems to work fine.

yonikid15: Oh true. Nice, I guess you could do the same for the `gov_email` to. 
`value="{{ user.gov_email ?: old('gov_email') }}"`

yonikid15: Wait, shouldn't Managers and HR Advisors only be hitting the registration form once. Or else they'd be able to change their department.

gggrant: 1. Done 👍 
2. They should be. More info on that in the issue description.

PR #2951, opened 2020-02-18
Add version to Job Applications

### LINKED ISSUE #2841: Add a version tag to Applications
Add a field to all applications that defines its version, probably a reference to a application_version status table. All existing applications must be set to 1, ones created with the timeline would be set to 2. This will allow us to keep using the old ui for old applications.

gggrant:
Resolves #2841.

### Notes:
- Do we want to tag everything v1 now, or later?

### DIFF
--- a/app/Models/JobApplication.php
+++ b/app/Models/JobApplication.php
@@ -75,7 +75,7 @@ class JobApplication extends BaseModel
         'submission_date' => 'string',
         'experience_saved' => 'boolean',
         'language_requirement_confirmed' => 'boolean',
-        'version_id'
+        'version_id' => 'int',
     ];
     protected $fillable = [
         'citizenship_declaration_id',
--- a/app/Models/JobApplication.php
+++ b/app/Models/JobApplication.php
@@ -7,12 +7,12 @@
 
 namespace App\Models;
 
+use App\Events\ApplicationRetrieved;
+use App\Events\ApplicationSaved;
 use App\Models\Applicant;
 use App\Models\ApplicationReview;
-use Illuminate\Notifications\Notifiable;
-use App\Events\ApplicationSaved;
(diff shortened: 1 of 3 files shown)


Khristophor:
I think we'll want to tag everything v1 right before we roll out the new Application flow, but I'm not sure how we'll remember to do that 😄

yonikid15:
> I think we'll want to tag everything v1 right before we roll out the new Application flow, but I'm not sure how we'll remember to do that 😄

Couldn't we just tag the applications v1 now? It wouldn't have any effect. Then when the new application flow comes out, make sure to set the version to v2.

gggrant:
> > I think we'll want to tag everything v1 right before we roll out the new Application flow, but I'm not sure how we'll remember to do that 😄
> 
> Couldn't we just tag the applications v1 now? It wouldn't have any effect. Then when the new application flow comes out, make sure to set the version to v2.

Just want to make sure no new applications would be missed during that in-between period of unknown length, for the sake of our live data.

app/Models/JobApplication.php
```diff
@@ -72,7 +74,8 @@ class JobApplication extends BaseModel
         'submission_signature' => 'string',
         'submission_date' => 'string',
         'experience_saved' => 'boolean',
-        'language_requirement_confirmed' => 'boolean'
+        'language_requirement_confirmed' => 'boolean',
+        'version_id'
```
yonikid15: Forgot to add the type here. Otherwise looks good 👍

yonikid15: Actually I can just add it in myself

Khristophor:
After thinking about it, we can add a migration to set the existing posters to v1, as well as some logic to set each application to the "current" version (we can set this in a config file or in the application model directly) and then update that version variable once we roll out the new application flow.

PR #2981, opened 2020-02-24
Experience models and database migrations

gggrant:
Resolves #2835.

- [x] Translations received for lookup tables.

PR #3018, opened 2020-02-28
Polymorphic n-n relationship between Skills-Experience

### LINKED ISSUE #2836: Experience-Skills relationship model and table
There needs to be an experience_skills lookup table. This needs to link our many kinds of experience experience with  skills, so if each experience type has its own table this may involve polymorphism.
Each entry also needs to contain a text field.

gggrant:
Resolves #2836.

### Notes:
- Merge #2981 first.


app/Models/Skill.php
```diff
@@ -1,14 +1,18 @@
...
 use Backpack\CRUD\app\Models\Traits\SpatieTranslatable\HasTranslations;
+use Illuminate\Database\Eloquent\Relations\Relation;
+
+Relation::morphMap([
+    'work' => 'App\Models\ExperienceWork',
+    'personal' => 'App\Models\ExperiencePersonal',
+    'education' => 'App\Models\ExperienceEducation',
+    'award' => 'App\Models\ExperienceAward',
+    'community' => 'App\Models\ExperienceCommunity',
+]);
```
tristan-orourke: This should be added to the morphMap in AppServiceProvider.
https://laravel.com/docs/7.x/eloquent-relationships#custom-polymorphic-types

app/Providers/AppServiceProvider.php
```diff
@@ -37,6 +42,11 @@ public function boot()
         Relation::morphMap([
             'applicant' => Applicant::class,
             'application' => JobApplication::class,
+            'work' => ExperienceWork::class,
+            'personal' => ExperiencePersonal::class,
+            'education' => ExperienceEducation::class,
+            'award' => ExperienceAward::class,
+            'community' => ExperienceCommunity::class,
```
tristan-orourke: Let's change these strings to 'experience_work', 'experience_personal', etc, to make the connection nice and explicit.

app/Models/Skill.php
```diff
@@ -73,14 +62,41 @@ public function classifications() // phpcs:ignore
...
+
+    public function experiences_award() // phpcs:ignore
+    {
+        return $this->morphedByMany(\App\Models\ExperienceAward::class, 'experience', 'experience_skills');
+    }
+
+    public function experiences_community() // phpcs:ignore
+    {
+        return $this->morphedByMany(\App\Models\ExperienceCommunity::class, 'experience', 'experience_skills');
+    }
```
tristan-orourke: This looks good!

app/Models/ExperienceAward.php
```diff
@@ -54,4 +55,9 @@ public function experienceable() //phpcs:ignore
     {
         return $this->morphTo();
     }
+
```
tristan-orourke: I think its useful to have the relationship to `skills`, but we also need to add a relationship to `experience_skills` on each of these Experience models, to get the justification.

Experience to ExperienceSkill is a one-to-many relationship, so I think the relationship would look like this:
```
public function experience_skills()
{
    return $this->morphMany(App\Models\ExperienceSkill::class, 'experience');
}
```

app/Models/ExperienceSkill.php
```diff
@@ -0,0 +1,64 @@
...
+
+    public function experiences_award() // phpcs:ignore
+    {
+        return $this->morphToMany(\App\Models\ExperienceAward::class, 'experience', 'experience_skills');
+    }
+
+    public function experiences_community() // phpcs:ignore
+    {
+        return $this->morphToMany(\App\Models\ExperienceCommunity::class, 'experience', 'experience_skills');
+    }
```
tristan-orourke: This is how the relationships need to be written from Skill to Experience (many-to-many) but Experience to ExperienceSkill is actually one-to-many . So I think we can get away with a single relationship here:
```
public function experience()
{
   return $this->morphTo();
}
```

app/Models/Skill.php
```diff
@@ -73,14 +62,41 @@ public function classifications() // phpcs:ignore
...
+
```
tristan-orourke: We may want to add a relationship to ExperienceSkill as well, for completeness. It would be a simple one-to-many relationship:
```
public function experience_skills()
    {
        return $this->hasMany(\App\Model\ExperienceSkill::class);
    }
```

app/Models/ExperienceSkill.php
```diff
@@ -0,0 +1,64 @@
...
+        'justification' => 'string',
+    ];
+
+    protected $fillable = [
+        'justification'
+    ];
+
+    public function skills()
+    {
+        return $this->belongsToMany(\App\Models\Skill::class);
```
tristan-orourke: Each ExperienceSkill corresponds to a single skill, so this should be `belongsTo` instead of `belongsToMany`.

PR #3019, opened 2020-02-28
Feature - Screen candidates email prompt

### LINKED ISSUE #2743: Manager / Prompt manager to screen candidates
# Design Goal & User Groups
We have observed that top candidates tend to drop out within 2 weeks of poster closing, we'd like to get managers start screening their candidates as soon as possible.

Rosita currently emails them to prompt them, but moving forward would like to automate this. prompt them directly within the app. 

So dev item for this issue:

1.  Send automated email to managers when a job poster closes. @nicolsonaitken please provide copy (translated ideally). Talent cloud should be CC'ed. Also CC any HR advisors that have claimed the job.

**Please select the users this UI will apply to.**
- [ ] Administrators
- [ ] [Applicants](https://github.com/GCTC-NTGC/TalentCloud/issues/1548)
- [x] [Hiring Managers](https://github.com/GCTC-NTGC/TalentCloud/issues/1549)
- [ ] [HR Advisers](https://github.com/GCTC-NTGC/TalentCloud/issues/1550)

yonikid15:
Resolves #2743 

### Notes:
- Sends email to managers when job poster closes prompting them to start screening applicants.

### Todo:
- [ ] Send email when job poster closes
- [ ] If job poster close date is updated, replace old email from the mail queue with updated one.


app/Listeners/SendScreenCandidatesEmail.php
```diff
@@ -0,0 +1,106 @@
...
+            'position_link' => [
+                'en' => route('jobs.show', $job),
+                'fr' => route('jobs.show', $job),
+            ],
+            'talent_cloud_email' => 'Talent.Cloud-nuage.de.talents@tbs-sct.gc.ca',
+        ];
+
+        $manager_email = $job->manager->user->email;
+        if (isset($manager_email)) {
+            Mail::to($manager_email)->send(new ScreenCandidatesPrompt($mailData));
```
tristan-orourke: Seems to me that, instead of assembling the data in this listener and passing the array to the Mailable class, we should just pass the job to the Mailable and let it extract its own data. This keeps data and its use situated more closely together. It would also make it easier to reuse the ScreenCandidatesPrompt elsewhere if we wanted to.

app/Listeners/SendScreenCandidatesEmail.php
```diff
@@ -0,0 +1,106 @@
...
+            ],
+            'num_of_applicants' => $num_of_applicants,
+            'num_of_noncitizens' => $num_of_noncitizens,
+            'num_of_veterans' => $num_of_veterans,
+            'position' => $job->getTranslations('title'),
+            'position_link' => [
+                'en' => route('jobs.show', $job),
+                'fr' => route('jobs.show', $job),
+            ],
+            'talent_cloud_email' => 'Talent.Cloud-nuage.de.talents@tbs-sct.gc.ca',
```
tristan-orourke: We have this email set in config/mail, you should reference that with `config('mail.from.address')`.

PR #3054, opened 2020-03-03
Bugfix - Prevent duplicate api requests

### LINKED ISSUE #2879: Task - Prevent multiple API requests to the same endpoint
# Description
I noticed on the HR Advisor job index, if every job displayed has the same manager, it makes multiple requests to the manager endpoint rather than making a single request and deferring to the content in the store for additional info.

https://talent.test/en/hr/jobs

# Images

yonikid15:
Resolves #2879

### DIFF
--- a/resources/assets/js/components/HRPortal/JobIndexHrPage.tsx
+++ b/resources/assets/js/components/HRPortal/JobIndexHrPage.tsx
@@ -25,10 +25,7 @@ import { RootState } from "../../store/store";
 import { fetchJobIndex } from "../../store/Job/jobActions";
 import { getAllJobs } from "../../store/Job/jobSelector";
 import RootContainer from "../RootContainer";
-import {
-  getManagers,
-  getManagerIsUpdatingById,
-} from "../../store/Manager/managerSelector";
+import { getManagers } from "../../store/Manager/managerSelector";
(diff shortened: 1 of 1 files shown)


resources/assets/js/components/HRPortal/JobIndexHrPage.tsx
```diff
@@ -256,15 +253,15 @@ const JobIndexHrDataFetcher: React.FC<JobIndexHrDataFetcherProps> = ({
...
-      find(managers, job.manager_id) === null &&
-      managersUpdating[job.manager_id] !== true
-    ) {
-      dispatch(fetchManager(job.manager_id));
-    }
-  });
+  useEffect(() => {
+    const uniqueManagers: number[] = [];
+    deptJobs.forEach((job: Job): void => {
+      if (!uniqueManagers.includes(job.manager_id)) {
```
tristan-orourke: I think this check can still include `find(managers, job.manager_id) === null`. No need to dispatch a fetch if we already have the data.

tristan-orourke: Well... We tried to change this but kept running into issues where modifying a hook dependency (ie `managers`) caused the whole loop to run again. We'll leave as is for now.

PR #3057, opened 2020-03-03
Add boolean to experience models for fulfilling education requirements

### LINKED ISSUE #2839: Add an "education_justification" field to Experience models
~Create a table of justification types. Currently FORMAL and COMBINATION. Create a lookup table between all experience types and the justification types. This means that any Experience object can be flagged as a FORMAL justification, a COMBINATION justification, both, or neither.~

All the new Experience models need a single boolean field to represent whether they satisfy education requirements or not.

gggrant:
Resolves #2839.

### Notes:
- Merge #2981 first.
- Default `true` for the education table?


database/migrations/2020_02_21_233331_add_experience_education_table.php
```diff
@@ -26,6 +26,7 @@ public function up()
             $table->boolean('has_blockcert')->nullable();
             $table->integer('experienceable_id')->unsigned()->index();
             $table->string('experienceable_type');
+            $table->boolean('is_education_requirement')->default(true);
```
tristan-orourke: Hmm. Good question, but I think we should leave it default(false) for consistency, and leave it up to the user. We can always suggest it in the UI if we want to.

app/Models/ExperienceAward.php
```diff
@@ -27,18 +28,20 @@ class ExperienceAward extends BaseModel
...
         'award_recipient_type_id',
         'issued_by',
         'award_recognition_type_id',
-        'awarded_date'
+        'awarded_date',
+        'is_education_requirement'
     ];
 
-    protected $table = 'experiences_award';
+    protected $table = ['experiences_award'];
```
tristan-orourke: Oh, I forgot one more comment! These table statements should be strings, not arrays (ie whats already in dev).

tristan-orourke: I'll fix it, and merge this in

gggrant:
> Normally its a bad idea to edit migrations created in a different PR, because its hard to know whether those migrations have been run on any of our servers yet (in which case the updated migration won't run the second time).
> If you're sure the experience table migrations haven't run anywhere yet, then it's okay this time.

Got it. I'm sure they haven't run in this case.

app/Models/ExperienceAward.php
tristan-orourke: Oh sweet, nvm.

PR #3124, opened 2020-03-13
Lift requirements on classification keys

### LINKED ISSUE #3117: Task - Lift requirements on classification keys
# Description
We currently restrict classifications to being 2-letter strings. We need to remove this restriction in order to allow a new classification: "EN-ENG". It can be combined with a classification level in the same way (eg EN-ENG-02).

gggrant:
Resolves #3117.


tristan-orourke:
Nice! I expected it to be more complicated.

PR #3175, opened 2020-03-17
Emergency Response - review applicants policy

### LINKED ISSUE #3151: Emergency Response - remove requirement that job must be closed before reviewing applicants
# Description
Conceptually, our emergency response job posters have no specific close date. Therefore it must be possible to review and sort applicants right away, while still accepting further applications.

If it is easy to allow this only for the new Emergency Response department, we should do so. However, it would be acceptable to turn off this requirement for all jobs temporarily, in which case we should create another git Issue reminding us to turn it back on.

yonikid15:
Resolves #3151 

### Notes:
- Assumes the `department->name == "Emergency Response`.

### DIFF
--- a/app/Models/JobPoster.php
+++ b/app/Models/JobPoster.php
@@ -95,6 +95,7 @@ use Spatie\Translatable\HasTranslations;
  * @method boolean isEditable()
  * @method string timeRemaining()
  * @method mixed[] toApiArray()
+ * @method boolean isInStrategicResponseDepartment()
  *
  * Computed Properties
  * @property string|null $classification_code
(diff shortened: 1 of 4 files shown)


app/Models/JobPoster.php
```diff
@@ -589,4 +589,15 @@ public function getClassificationMessageAttribute()
...
+
+    /**
+     *
+     * Job Poster is from the Emergency Response department
+     *
+     * @return boolean
+     */
+    public function isInEmergencyResponseDepartment()
+    {
+        return $this->department->name == 'Emergency Response';
```
tristan-orourke: Careful here. There's a few issues with checking the name. First, if the request was in french `->name` will be different. Also,  admins could change the name on the Admin portal without telling us.

Checking against any hardcoded value is way less than ideal, but lets use the id instead, as its less likely to change. The id of this department on production is **18**.

yonikid15: Ya I wasn't sure about this yet. Thanks.

PR #3211, opened 2020-03-20
Homepage message

joshdrink:
Adds message to homepage for maintenance


resources/views/applicant/home.html.twig
```diff
@@ -42,6 +42,7 @@
 			{# Browse Jobs ================================================== #}
 			<div class="home-browse" data-c-background="white(100)" data-c-padding="tb(3)" data-c-border="bottom(thin, solid, grey)">
 				{% set christmas_time = false %}
+                {% set emergencyResponse = true %}
```
tristan-orourke: These should really be set in the controller. I`ll make the change.

tristan-orourke:
I moved the boolean switches for Christmas and COVID messages to env variables, to let us turn them on and off without a full deployent.

PR #3238, opened 2020-03-23
Changes to Application for Strategic Talent Response jobs

### LINKED ISSUE #3167: Emergency Response - Changes to Application
#Description
We want to hide certain fields and sections of the application (to Emergency Response dept jobs) in order to simplify it. 

The tricky part is that this may also necessitate changes to application validation. We may be able to get around this by setting default values in hidden fields.
edit: None of these fields are required in the database, so we CAN remove the fields entirely and use a different ApplicationValidator.

Definitely hide/change:
[ ] Simplified copy on the Experience step
[ ] Hide Courses/Certificates and Alternate Experience (ie only show Degrees on Step 2)
[ ] Hide all Soft Skills that that are also Nice to Have

Probably hide/change:
[ ] Hide Citizenship status
[ ] Hide Veteran status
[ ] Hide Language requirement? (This can probably stay if we set it to English or French)

Nice to hide/change:
[ ] Hide Skill Level on skill declarations
[ ] "Status: claimed" on skill declarations
[ ] Hide references & work samples on skill declarations

tristan-orourke:
Resolves #3167.

### Notes:
- I copied the entire folder of application twig templates to create a parallel set of STR specific views.
- Mostly, the new views have certain portions removed (some basic info fields, degrees & work experience, skill levels)
- A new Application validator is used for STR jobs, to allow for the missing fields
- I removed the requirement for skill levels for ALL skill declarations, because that was easiest
- I did some refactoring of the twig templates while I was mucking around. The STR application now uses separate templates for the skill form, and the skill view on application previews. (The normal application still uses `common/skill.html.twig`, a mass of conditional statements.)


app/Http/Controllers/ApplicationByJobController.php
```diff
@@ -91,15 +93,19 @@ public function editBasics(JobPoster $jobPoster)
         $this->authorize('view', $application);
         $this->authorize('update', $application);
 
+        $viewTemplate = $jobPoster->isInStrategicResponseDepartment()
```
Khristophor: I probably would have done something like
```php
$templatePath = $jobPoster->isInStrategicResponseDepartment() ? "/strategic_response_application" : "";
...
return view (
    "applicant$templatePath/application_post_01",
```
Just to avoid checking/possible typos on every render, but this works fine as-is.

PR #3315, opened 2020-03-30
Add Response Screening page

Khristophor:
## OOOOOOKAY FOLKS

Heading to `manager/jobs/{id}/applications` for a Job Poster with the Department === "Strategic Talent Response" (currently `id: 18`) will load the new screening UI.

#### This has been built from the ground up with the following features:

- Redux store for Applications and Departments
- A separate Formik instance for each Application row
- Sweet alerts for note edits and actions with side effects (allocated, not available)
- Confirmation dialog when trying to leave a page with unsaved changes (this is browser native, and not configurable)
- Applicant sorting within buckets:
  - Unreviewed, ready for reference check, and further assessment required appear in "Under Consideration" (in that order)
  - Allocated are sorted by department not yet set, and department name
  - Every other status is 1:1 with a bucket
- When setting an applicant to Allocated and saving, an alert should appear warning the user that this action will affect all other applications. The controller then runs an update to all other relevant (submitted) applications if the poster is in the Strategic Talent Response department.
- When setting an applicant to Unavailable and saving, an alert should appear warning the user that this action will affect all other applications. The controller then runs an update to all other relevant (submitted) applications if the poster is in the Strategic Talent Response department.
- Removed the individual application review component from the application view for a Strategic Talent Response job

#### Things left to do:

- Translations

I also haven't really written tests, so there's that.


app/Http/Controllers/Api/ApplicationController.php
```diff
@@ -0,0 +1,129 @@
...
+     * belonging to a given Applicant. Designed to be used for the
+     * Strategic Talent Response screening, but could be repurposed.
+     *
+     * @param JobApplication $application Incoming Job Application object.
+     *
+     * @return void
+     */
+    protected function setAvailability(JobApplication $application)
+    {
+        $departmentId = $application->job_poster->department_id;
```
tristan-orourke, line 113: I really like how you implemented this whole function! 👍

app/Http/Controllers/Api/DepartmentController.php
```diff
@@ -15,6 +15,8 @@ class DepartmentController extends Controller
      */
     public function index()
     {
-        return JsonResource::collection(Department::all());
+        $strategicDepartmentId = config('app.strategic_response_department_id');
+        $departments = Department::where('id', '<>', $strategicDepartmentId)->get();
+        return JsonResource::collection($departments);
```
tristan-orourke, line 20: Won't this make it impossible for new hr advisors to register for this department?

resources/assets/js/components/StrategicTalentResponse/ResponseScreening/ApplicantBucket.tsx
```diff
@@ -0,0 +1,684 @@
...
+    />
+  );
+};
+
+// Kinda weird "empty" component that hooks into Formik's
+// context, listens to the 'dirty' prop, and registers
+// a beforeunload listener to fire if a user attempts to
+// leave with unsaved work.
+// https://github.com/jaredpalmer/formik/issues/1657#issuecomment-509388871
+const AlertWhenUnsaved = (): React.ReactElement => {
```
tristan-orourke, line 155: This seems useful enough to turn into our own hook.

app/Http/Controllers/Api/DepartmentController.php
tristan-orourke: I suppose admins can still set it before upgrading the user to the HrAdvisor role.

PR #3340, opened 2020-03-31
Bugfix - Strategic Response login/logout/register redirect

### LINKED ISSUE #3316: Bug - Redirect to Strategic Response homepage after login/logout
# Description
When user tries to login/logout/register it redirects them to the applicant homepage instead of the strategic response landing page. 

# Steps to Reproduce
1. Go to Response homepage. 
2. Click on Login and then login
3. See error

yonikid15:
Resolves #3316

### DIFF
--- a/app/Http/Controllers/Auth/LoginController.php
+++ b/app/Http/Controllers/Auth/LoginController.php
@@ -51,6 +51,13 @@ class LoginController extends AuthController
         } else {
             $redirectTo = route('home');
         }
+
+        // Redirect to Response landing page if user logged in from that page.
+        $previous_url = session()->get('response.index');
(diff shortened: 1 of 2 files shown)


tristan-orourke:
> This looks good in terms of what we have, but I don't really understand why we aren't just redirecting to whatever the previous url was in all cases during login/logout. It seems like we have a ton of conditional logic around a fairly straightforward convention. Any info on this @tristan-orourke ?

Yeah, I don't really remember why we aren't doing that already. The only issue I can think of is if you navigate manually to `/login`, "back" could send you off the site entirely. That's very much a corner case though. We could ignore it entirely, or send you to `route("home")` if the previous url is a different domain.

PR #3501, opened 2020-04-23
Task - refactor breadcrumbs

### LINKED ISSUE #2699: Bug - Job Related Breadcrumbs Showing ID
# Description
It would be more useful for the end user to have the job title in the breadcrumb list instead of the job's ID, because the ID means very little to the user.

# Technical Info

**Device:** All
**Operating System:** All
**Browser:** All

# Steps to Reproduce
1. Visit any job related page with breadcrumbs
2. See ID in breadcrumbs

yonikid15:
Resolves #2699 

### Notes:
- I switched the Job Poster Preview page url with the Job Summary page url (the applicant side keeps the same url to the poster).
- I have made some url changes to utilize the implicit binding feature on laravel, allowing to replace ids with title/names.
    - `'managers/{manager}'` -> `'jobs/{jobPoster}/manager/{manager}'`
    - `'applications/{application}'` -> `'jobs/{jobPoster}/applications/{application}'`
    - `'applicants/{applicant}'` -> `'jobs/{jobPoster}/applicants/{applicant}'`

### Update: 

- Fixed browser tests


resources/views/common/breadcrumbs.html.twig
```diff
@@ -1,19 +1,19 @@
 <nav aria-label="breadcrumb" data-c-breadcrumbs data-c-colour="{{ font_color ?: 'white' }}">
-	<ol>
-		{% for key, url in breadcrumbs %}
+	<ol id="{{ breadcrumbs_id }}">
```
tristan-orourke, line 2: From what I can see this is not set on most pages, resulting in an empty id tag which is technically invalid HTML.

resources/views/common/breadcrumbs.html.twig
```diff
@@ -1,19 +1,19 @@
...
+		{% for key, url in custom_breadcrumbs ?: breadcrumbs %}
 			{% if loop.first %}
 				<li>
 					<a href="{{ url(url) }}">{{ breadcrumbs_lang.home }}</a>
 					<i class="fas fa-caret-right"></i>
 				</li>
```
tristan-orourke, line 8: Why does the first item in the loop need any special behaviour?

routes/web.php
```diff
@@ -93,7 +93,8 @@ function (): void {
             Route::middleware(['auth'])->group(function (): void {
 
                 /* Managers */
-                Route::get('managers/{manager}', 'ManagerProfileController@show')
+                Route::get('jobs/{jobPoster}/manager/{manager}', 'ManagerProfileController@show')
```
tristan-orourke, line 96: If we're using the job id in the url, we actually don't need the manager id anymore, since each job has a unique manager.

routes/web.php
```diff
@@ -348,27 +349,29 @@ function (): void {
...
-                        Route::get('applications/{application}', 'ApplicationController@show')
+                        Route::get('jobs/{jobPoster}/applications/{application}', 'ApplicationController@show')
+                            ->middleware('can:manage,jobPoster')
                             ->middleware('can:view,application')
                             ->name('manager.applications.show');
 
                         /* View Applicant Profile */
-                        Route::get('applicants/{applicant}', 'ApplicantProfileController@show')
+                        Route::get('jobs/{jobPoster}/applicants/{applicant}', 'ApplicantProfileController@show')
+                            ->middleware('can:manage,jobPoster')
```
tristan-orourke, line 359: I was unsure about this, since the jobPoster id isn't technically necessary, but after talking to the designers I see why this would make more sense to users.

resources/views/common/breadcrumbs.html.twig
yonikid15: The breadcrumbs are extracted and built using the URL segments. So the first key in the segments array, is always the locale, homepage (eg. `en => en`). Since, the key is used in the breadcrumbs, this replaces this with the word "Home". I guess I could add it to the list of keys in the lang files if you think that's preferable.

tristan-orourke: Ah, no that makes sense. Maybe just add a comment: "Replace en/fr with a link to Home."

app/Http/Controllers/ManagerProfileController.php
```diff
@@ -21,10 +23,11 @@ class ManagerProfileController extends Controller
      * Show a manager profile
      *
      * @param  \Illuminate\Http\Request $request Incoming Request.
+     * @param  \App\Models\JobPoster    $jobPoster Incoming JobPoster object.
      * @param  \App\Models\Manager      $manager Incoming Manager.
      * @return \Illuminate\Http\Response
      */
-    public function show(Request $request, Manager $manager)
+    public function show(Request $request, JobPoster $jobPoster, Manager $manager)
```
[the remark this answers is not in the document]
yonikid15: Umm possibly. Job posters always have a manager though so this should work for now. All links on the site that lead to managers are connected through a job poster (I believe).

resources/views/common/header.html.twig
```diff
@@ -14,10 +14,13 @@
             {% endif %}
             <div data-c-margin="top(2)">
                 <div data-c-grid="gutter(all, 1)">
-                    <div data-c-grid-item="tl(1of2)">
-                        {% include "common/breadcrumbs" %}
+                    <div data-c-grid-item="tl(3of4)">
+                        {% if hide_breadcrumbs %}
+                        {% else %}
+                            {% include "common/breadcrumbs" %}
```
tristan-orourke, line 20: This line is throwing an odd error when I visit the Job Builder review page.

resources/assets/js/helpers/router.tsx
```diff
@@ -45,20 +52,65 @@ export const useRouter = (
...
+
+    icon.classList.add("fas", "fa-caret-right");
+    breadcrumb.append(anchor);
+    breadcrumb.append(icon);
+
+    if (breadcrumbs) {
+      const lastBreadcrumb = breadcrumbs.lastElementChild;
+      breadcrumbs.insertBefore(breadcrumb, lastBreadcrumb);
+    }
+  };
```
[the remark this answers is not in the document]
tristan-orourke: I still don't like all this JPB-specific code in the router hook. I'll think about how to reorganize it.

Issue #3472, opened 2020-04-20
Bug - Emergency Response - Header of Applicant Sorting page is the same regardless of job


tristan-orourke:
Reported by @rositak: 
> I just noticed that on the sorting page, the information in the header "You are reviewing applicants for: Policy Analysis - Economic Programs - Senior Level." is the same regardless of the actual position

> I've looked at the corporate services stream, the Admin stream, and one of the IT stream, and they are all the same

PR #3557, opened 2020-04-30
Bugfix - Static Response Screening header

Khristophor:
#3472 
Fixed up the hardcoded bits in the Twig header, needs French translations to the newly added langfile.


tristan-orourke:
Oh, wait. These translations fail the phpunit tests. Maybe we should wait for the french, to merge.

PR #3559, opened 2020-04-30
Bugfix - Blank JPB review page for new manager accounts

Khristophor:
Added another check on the `localizeField` function in the event `model[field]` is null (which was the case here).

### DIFF
--- a/resources/assets/js/helpers/localize.ts
+++ b/resources/assets/js/helpers/localize.ts
@@ -13,7 +13,7 @@ export function localizeField<T>(
   model: T,
   field: TranslatableKeys<T>,
 ): string | null {
-  if (model[field] && model[field][locale]) {
+  if (model[field] !== null) {
     return model[field][locale];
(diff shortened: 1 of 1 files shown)


resources/assets/js/helpers/localize.ts
```diff
@@ -13,7 +13,10 @@ export function localizeField<T>(
   model: T,
   field: TranslatableKeys<T>,
 ): string | null {
-  return model[field][locale];
+  if (model[field] && model[field][locale]) {
```
tristan-orourke, line 16: I'd do an explicit `!== undefined` check, and not bother to check `model[field][locale]`. As it is, if `model[field][locale]` is the empty string it will evaluate as falsey and will return null instead.

Khristophor: Added a `!== null` check, `undefined` was still failing and throwing an error.

PR #3659, opened 2020-05-14
Resolve npm audit issues

Khristophor:
### Notes:
- The removal of `dev: true` throughout package-lock.json seems to be an [ongoing npm bug](https://npm.community/t/package-lock-json-keeps-changing-between-platforms-and-runs/1129/17)
- storybook addon info is apparently deprecated, and was throwing a vulnerability so I removed it for now. The alternative is [storybook addon docs](https://github.com/storybookjs/storybook/blob/next/addons/docs/README.md)

### DIFF
--- a/.storybook/config.ts
+++ b/.storybook/config.ts
@@ -1,16 +1,10 @@
 import React from "react";
 import { addParameters, configure, addDecorator } from "@storybook/react";
 import { themes } from "@storybook/theming";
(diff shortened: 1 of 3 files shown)


Khristophor:
> The withInfo decorator is still being added in .storybook/config.ts
> 
> I'm surprised this is deprecated, the last commit was only a month ago!

Good catch! I'm also surprised, but the docs addon seems a lot more robust.

PR #3840, opened 2020-06-04
Applicant Timeline - Welcome Step

Khristophor:
### Notes:
- French translations required
- Zero functionality, simply has text and a button with a passable click handler


Khristophor:
> Looks good 👍
> 
> However, should decide on a naming convention for the formatted message id's and the folder structure. Chris is using different names then I on the Basic Info PR.
> 
> Example for formatted message:
> 
> `id: "applicationTimeline.{name of the step}.header"`
> OR
> `id: "applicantTimeline.{name of step}.header"`
> 
> Example for folder structure.
> 
> `Application/files OR `ApplicantTimeline/{name of step}/files for that step`
> 
> @tristan-orourke @Khristophor

Agreed, we do need a convention. I had copied @tristan-orourke 's convention for file naming, but I didn't really pay attention to the lang id. Maybe stick to the component or folder name, so `application.[step name].header` or similar?

tristan-orourke:
Re: naming conventions
I think each step should have its own folder, since some of the steps already have many components. (eg the experience step with accordion components for each experience type.) And just `Application` is fine; we've mostly use the term `ApplicationTimeline` to distinguish it from our previous version, but the previous version has no React files anyway.

So that looks like `Application/{name of step}/files for that step`.

And I like `application.[step name].header` for message ids, since that matches the file convention. Maybe `application.[component name].header` for sub-components like accordions.

Khristophor:
@yonikid15 @tristan-orourke updated files/keys to the convention.

PR #3849, opened 2020-06-05
Contact Preferences Changes

bitliege:
Okay so there are a couple UI things that have changed here that will require dev work - unfortunately I didn't have time to look at translations, but I'm happy to do them later if you guys want me to take care of them.

Registration:
- there are two new questions that are associated with the users account that they must now answer:
  - your preferred contact language
  - an opt-OUT checkbox to indicate they do not wish to be contacted about related jobs

Account Settings:
- these same two questions exist here so that the user can change them

New Application Basic Info Step:
- these questions have been removed

New Application Review Step:
- there is a new section here to reflect the user's contact preferences in their profile
- there is a new question at the bottom of this page that needs dev work where the user indicates whether they would like this manager to share their application with other managers looking for similar skill sets

All application steps now have a new div with a border to better divide the page content from the footer too!

## What's left
- [ ] still needs translations


gobyrne:
@yonikid15 let's send the translations for this to @nicolsonaitken .

PR #3880, opened 2020-06-11
Add API endpoint for getting and updating basic application info

Khristophor:
### Notes:

- Added a couple confirmation fields for the UI: `language_test_confirmed` and `education_requirement_confirmed`
- Added a new Resource class for returning basic Application info... I waffled a bit between using the existing Application Resource but this seemed an easier way to control what fields are returned
- Added a form validator for posted fields, this doesn't check the validity in terms of business rules (i.e. that an applicant is legally able to work in Canada), just that the fields are present and are of the correct data type

### DIFF
--- a/app/Http/Controllers/Api/ApplicationController.php
+++ b/app/Http/Controllers/Api/ApplicationController.php
@@ -3,7 +3,9 @@
 namespace App\Http\Controllers\Api;
 
 use App\Http\Controllers\Controller;
+use App\Http\Requests\UpdateJobApplicationBasic;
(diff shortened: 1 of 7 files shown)


Khristophor:
> Looks good 👍
> However, I believe the `preferred_language_id` is being moved to the Registration form (and can be changed by the user on the Settings page).
> 
> Update:
> Just confirmed with Josh that this is the case. So the `preferred_language_id` will not be part of the Basic Info step anymore.

@yonikid15 Took out preferred language stuffs from both endpoints.

PR #3912, opened 2020-06-18
Add EN-ENG classification to constants

Khristophor:


tristan-orourke:
Should we add a seeder that adds the extra classification, to keep our dev environments in line?

tristan-orourke:
Though, if we connect it to the DB properly soon, a seeder is less useful.

PR #4009, opened 2020-07-03
Ensure localizeFieldNonNull doesn't cause an exception if a localization is missing

tristan-orourke:
Resolves an issue Rosita ran into where the JPB Skill step crashed when new skills had recently been created, with french text not added yet.


tristan-orourke:
Actually, localizeField and localizeFieldNonNull have to be able to return an empty string. There are times when that will be the indended value.

PR #4013, opened 2020-07-06
Application Timeline - My Fit step (Long-form questions)

### LINKED ISSUE #2888: Task - Application - UI for Long-form Questions step of the Application
# Description
Create UI in react for filling out any number of long-form application questions. Each answer textarea should include a word-counter component.

Though the wireframe includes a Skill Questions section with an associated skill, we can ignore this for now.

# Images
https://xd.adobe.com/view/a678f763-f7b7-43ca-740a-89359cbe42f1-d049/screen/3bd349c0-6dff-49ae-abad-70b1d27c0f43/Step-2B-Requirements-7

yonikid15:
Resolves #2888 

### Notes:
- To accomplish saving all questions individually as well as having a "Submit All" button I used reacts reference hooks. Formik comes with a `innerRef` prop for that purpose. This allows triggering submission (as well as access to other Formik props) outside the form. 
- Currently, the "Save All" button triggers the submission on all the individual forms. It's probably better to create an array of all the answers and make a single request instead. This shouldn't be hard to implement.


resources/assets/js/models/types.ts
```diff
@@ -172,12 +172,26 @@ export interface Job {
...
   id: number;
   job_poster_id: number;
   description: localizedFieldNonNull;
 }
 
+export interface JobPosterQuestion {
+  id: number;
+  job_poster_id: number;
+  question: localizedFieldNonNull;
+  description: localizedFieldNonNull;
```
tristan-orourke, line 192: These fields ARE nullable, so use `localizedField` instead. Generally, most content entered by a user (including a manager) can be null. `localizedFieldNonNull` should mostly be used for constants that are only created by devs or admins, like Skills or Provinces.

resources/assets/js/components/Application/MyFit/MyFit.tsx
```diff
@@ -0,0 +1,208 @@
...
+          id="application.myfit.firstParagraph"
+          defaultMessage="The manager has included a few further questions that they'd like you to answer in order to be considered for this position. Questions that require an answer are marked, while all others are optional."
+          description="Paragraph for the My Fit step in the Application Timeline"
+        />
+      </p>
+      {jobQuestions.map((question, index) => {
+        formRefs.current = [
+          ...formRefs.current,
+          formRefs.current[index] || createRef(),
+        ];
```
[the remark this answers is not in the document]
yonikid15: Yup! I saw that during review yesterday and made a fix.

resources/assets/js/components/Application/MyFit/MyFit.tsx
```diff
@@ -0,0 +1,208 @@
...
+        ref.current.submitForm(),
+      ),
+    );
+  };
+
+  const validateAndSubmit = (
+    refs: React.MutableRefObject<
+      React.MutableRefObject<FormikProps<QuestionValues>>[]
+    >,
+    nextStep: string,
```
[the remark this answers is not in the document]
yonikid15: This is supposed to be `nextStepUrl` whenever we establish what the routes will be.

resources/assets/js/components/Application/MyFit/MyFit.tsx
```diff
@@ -117,10 +116,12 @@ export const MyFit: React.FunctionComponent<MyFitProps> = ({
...
-        formRefs.current = [
-          ...formRefs.current,
-          formRefs.current[index] || createRef(),
-        ];
+        if (!formRefs.current.includes(formRefs.current[index])) {
+          formRefs.current = [
+            ...formRefs.current,
+            formRefs.current[index] || createRef(),
+          ];
+        }
```
tristan-orourke, line 69: This doesn't seem quite right. The if statement asks is `current` includes an element of `current` - looks like circular logic. And I'm not sure its safe to use index as the key. If a question is deleted, the same index may later refer to a different question. Maybe using question.id would be better?

resources/assets/js/helpers/forms.ts
```diff
@@ -0,0 +1,47 @@
+import { FormikProps, FormikValues } from "formik";
+import _ from "lodash";
```
tristan-orourke, line 2: You can import only the function you need for a smaller file, like this: `import isEmpty from "lodash/isEmpty";`

resources/assets/js/models/types.ts
```diff
@@ -172,12 +172,26 @@ export interface Job {
   culture_special: localizedField;
 }
 
+export interface JobApplicationAnswer {
+  id: number;
+  job_poster_questions_id: number;
+  job_application_id: number;
+  answer: localizedField;
```
tristan-orourke, line 179: Should be just a string.

resources/assets/js/components/Application/MyFit/Question.tsx
```diff
@@ -0,0 +1,169 @@
...
+interface QuestionProps {
+  jobApplicationAnswer: JobApplicationAnswer;
+  formRef: any;
+  index: number;
+  question: JobPosterQuestion;
+  handleSubmit: (data: JobApplicationAnswer) => Promise<void>;
+}
+
+export interface QuestionValues {
+  id: number;
```
tristan-orourke, line 34: If this is just the Form Values, I think the only property necessary here is the `answer` field.

resources/assets/js/helpers/forms.ts
```diff
@@ -0,0 +1,47 @@
...
+
+export const focusOnElement = (id: string): void => {
+  const element = document.getElementById(id);
+  if (element) {
+    element.focus();
+  }
+};
+
+export const validateAllForms = async (
+  refs: MutableRefObject<MutableRefObject<FormikProps<FormikValues>>[]>,
```
tristan-orourke, line 13: I realized while poking around that in this function and submitAllForms, we just need the inner array of refs, there's no need to pass in the array wrapped in another MutableRefObject.

yonikid15: Noted. Thanks again for the help. I'm going to keep looking for a solution to the `focusOnElement` issue tomorrow.

resources/assets/js/components/Application/MyFit/MyFit.tsx
```diff
@@ -116,11 +115,8 @@ export const MyFit: React.FunctionComponent<MyFitProps> = ({
               data-c-radius="rounded"
               type="button"
               disabled={isSubmitting}
-              onClick={() =>
-                validateAndSubmit(
-                  formRefs.current,
-                  "ReplaceWithUrlOfPreviousStep",
-                )
+              onClick={(): Promise<void> =>
+                validateAndSubmit(formRefs.current).then(handleReturn)
```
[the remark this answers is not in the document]
yonikid15: This was actually your suggestion on friday :p I just forgot to push it up.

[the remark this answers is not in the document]
tristan-orourke: Aw, after testing it this doesn't quite work yet. First of all, the handle______ functions trigger even if validation or submission fails. But also, the handle____ functions are triggering _before_ the handleSubmit callback completes.

PR #4031, opened 2020-07-09
Task - Confirmation modal for Job Poster status transition

### LINKED ISSUE #3875: Manager/HR - Confirmation Model for Job Poster status transitions
Add a confirmation model for all job poster transitions triggered on the Summary Page.

Text should look like this:

**[Transition Name]?**
This will change the job poster from [From Status] to [To Status].

**Cancel**                                                                           **Confirm** 

For example:
**Send to Translation?**
This will change the job poster from **In Review (HR)** to **In Translation** 
**Cancel**                                                                           **Confirm (edited)**

yonikid15:
Resolves #3875 

### Notes:
- Added a confirmation popup for relinquishing a job as well.

### DIFF
--- a/app/Http/Controllers/JobSummaryController.php
+++ b/app/Http/Controllers/JobSummaryController.php
@@ -80,6 +80,8 @@ class JobSummaryController extends Controller
         $transitionToButton = function (JobPosterStatusTransition $transition) use ($user, $jobPoster, $transitionManager) {
             return [
                 'text' => $transition->name,
+                'from_status' => $transition->from->name,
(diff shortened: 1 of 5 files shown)


tristan-orourke:
@gobyrne Do we want to add the description of the next status (the one you're transitioning to) to the confirmation modal? Or are you happy with it as seen in the sprint review demo?

gobyrne:
Leave as is for now. We think that with the descriptions that will be added to the summary page, this will be sufficient. We can always add it later if we have to.

PR #4086, opened 2020-07-16
Update deployment scripts

### LINKED ISSUE #4059: Bug - Files from "Resources" page missing (on admin portal too)
# Description
We created a resources page where admins could upload files (like email templates) for managers and HR to use. The original files that were uploaded are now missing. We can upload them again, but wanted to check in first.

tristan-orourke:
Resolves #4059.

### Notes:
- I created a single script that accepts two parameters: the name of the zip file containing new code, and the destination app folder. This will make it easier to maintain, because there's no more duplication of code between different deploy scripts. For convenience, I set up separate scripts for each situation (prod/dev and tc/reserve) that call the main script with appropriate arguments.
- This new script still needs to be tested (on dev), since I don't a working tablet right now.

### DIFF
--- a/etc/scripts/deploy-instructions.txt
+++ b/etc/scripts/deploy-instructions.txt
@@ -5,8 +5,17 @@ composer i --no-dev --optimize-autoloader && npm run prod && npm ci --production
 
 3. Zip the TalentCloud folder, copy to server on TBS device.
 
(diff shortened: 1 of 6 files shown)


tristan-orourke:
I'm tested this on dev and prod, the new upload scripts work correctly.

PR #4276, opened 2020-08-24
Application Timeline - Review UI

Khristophor:
### Notes:
- The "share application" at the bottom needs a mini form with somewhere to actually store that data (application migration)
- Some UI states are still weird/unaccounted for, things like
  - No experiences attached to a required skill in the skill view (not sure if this is possible, but the mock data can allow it)
- Links to other sections (the edit urls)


resources/assets/js/components/Application/Review/SkillAccordion.tsx
```diff
@@ -0,0 +1,732 @@
...
+            return (
+              <ExperienceContent
+                key={`${experience.type}=${experience.id}-skill`}
+                intl={intl}
+                locale={locale}
+                experience={experience}
+                justification={justification}
+              />
+            );
+          })}
```
yonikid15, line 725: If the experiences length is zero, we should probably have a message in the accordion telling the applicant they have no experiences attached to this skill.

PR #4296, opened 2020-08-26
Miscellaneous things for Azure. The pipeline cometh!

gggrant:
Should be okay to merge any Azure changes into `dev`, then setup `dev` as the continuous deployment branch.

### Notes:
- The pipeline cometh!


gggrant:
Opened to test `azure-pipelines.yml` functionality, which won't actually be utilized for our use case.

Releases are triggered when pushing to Azure remote repo.

PR #4342, opened 2020-09-02
Application - Make Skill step fully functional

### LINKED ISSUE #2869: Applicant - Make Skill Information step fully functional
# Description
Making the step functional means tying it into the redux store and APIs so that it get can GET and UPDATE application data, as well as navigate to other pages (SAVE AND EXIT or CONTINUE).

This page also requires some sort of page-level flag to keep track of unsaved changes, that could trigger an alert when a user attempts to navigate away from the page if they have content in any field that is unsaved.

# Images
https://xd.adobe.com/view/a678f763-f7b7-43ca-740a-89359cbe42f1-d049/screen/9fe5fb8a-ee83-4da0-8ca8-6c520f579600/

tristan-orourke:
Resolves #2869.

This is branched off the Experience Step PR (#4341) due to much identical code. Ensure that PR is merged first.

Notes
- I added the skills intro page as well. It starts fetching all the data needed for the Skill step, so that repeated code should be extracted, but we can do that when all the steps are finished, as we expect a fair bit of common code between them.


app/Http/Controllers/ApplicationTimelineController.php
```diff
@@ -12,6 +12,7 @@ public function show(JobApplication $jobApplication)
         return view('applicant/application-timeline-root')
             ->with([
                 'title' => $jobApplication->job_poster->title, // TODO: Check with design what the title should be.
+                'diable_clone_js' => true,
```
Khristophor, line 15: This is typo'd, so it shouldn't actually have fixed anything 🤔

tristan-orourke: It seems it was an intermittent issue in the first place, and I just got lucky after adding this typo'd line 😅 . Fixing the typo _definitely_ helped.

yonikid15 (changes requested):
Great job 👍 Just a few issues/questions:

- Closing the accordion without hitting save scraps the text the user just entered. It's cause when the accordion is collapsed the entire element is removed from the page. Well need to find a way to be able collapse the accordion without deleting the contents of the form inside. 
- Validation using the navigation buttons only works if the accordions are open. Again, this is due to the element getting removed when collapsed.
- Shouldn't the status icons get validated on when the page first opens? Right now you need to submit them to cause a change.

resources/assets/js/components/Application/Skills/skillsHelpers.ts
```diff
@@ -71,6 +72,19 @@ export const computeParentStatus = (
   return IconStatus.DEFAULT;
 };
 
+/** Get current experienceSkill status if it's stored in status store, or return DEFAULT. */
+export const computeEperienceStatus = (
```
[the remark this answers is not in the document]
tristan-orourke: 👍

tristan-orourke:
> Great job 👍 Just a few issues/questions:
> 
> * Closing the accordion without hitting save scraps the text the user just entered. It's cause when the accordion is collapsed the entire element is removed from the page. Well need to find a way to be able collapse the accordion without deleting the contents of the form inside.
> * Validation using the navigation buttons only works if the accordions are open. Again, this is due to the element getting removed when collapsed.
> * Shouldn't the status icons get validated on when the page first opens? Right now you need to submit them to cause a change.

All these have been fixed, I think.

resources/assets/js/components/Application/Skills/Skills.tsx
```diff
@@ -64,7 +82,7 @@ const Sidebar: React.FC<SidebarProps> = ({ menuSkills, intl, status }) => {
       </p>
       <ul>
         {Object.keys(menuSkills).map((skillId) => (
-          <li key={skillId}>
+          <li key={`application-skills-sidebar-skill-${skillId}`}>
```
Khristophor, line 85: I don't know that it's necessary to be so verbose with keys, see https://reactjs.org/docs/lists-and-keys.html#keys-must-only-be-unique-among-siblings

resources/assets/js/components/Application/Skills/Skills.tsx
```diff
@@ -249,129 +271,125 @@ const ExperienceSkillAccordion: React.FC<ExperienceSkillAccordionProps> = ({
...
+            <Form>
+              <AlertWhenUnsaved />
+              <hr data-c-hr="thin(gray)" data-c-margin="bottom(1)" />
+              <div data-c-padding="lr(1)">
+                <FastField
+                  id={`experience-skill-textarea-${experienceSkill.id}`}
+                  name="justification"
+                  label={label}
+                  component={TextAreaInput}
+                  placeholder="Start writing here..."
```
Khristophor, line 318: Placeholder text should be localized

resources/assets/js/components/Application/Skills/Skills.tsx
tristan-orourke: Yeah, you're right. I think I was confused about what I was trying to fix when I did this :P I'll reverse it.

Khristophor (commented):
Nice work 👍 

Just need to update `skillsHelpers.test.ts` to match the new initial state for the status icons.

yonikid15 (approved):
Looks good 👍

PR #4353, opened 2020-09-04
Application - Make Basic Info step fully functional

Khristophor:
### Notes:
Check out `https://talent.test/en/demo/applications/{applicationId}/basic` to test out locally


resources/assets/js/components/Application/BasicInfo/BasicInfoPage.tsx
```diff
@@ -1,59 +1,121 @@
...
+  useEffect(() => {
+    // If job is null and not already updating, fetch it.
+    if (jobId && job === null && !jobIsUpdating) {
+      dispatch(fetchJob(jobId));
+    }
+  }, [jobId, job, jobIsUpdating, dispatch]);
+
+  if (application === null || job === null) {
+    return null;
+  }
```
tristan-orourke, line 73: Is returning null really how we want to handle this? We could render the progress bar before these are loaded, and show some kind of loading state...

resources/assets/js/components/Application/BasicInfo/BasicInfoPage.tsx
```diff
@@ -1,59 +1,121 @@
...
+    if (!result.error) {
+      const payload = await result.payload;
+      return payload;
+    }
+    return Promise.reject(result.payload);
+  };
+
+  const handleContinue = (values: ApplicationNormalized): void => {
+    updateApplication(values);
     navigate(applicationExperienceIntro(locale, applicationId));
```
tristan-orourke, line 88: As is, this will start the update but navigate away before it completes (and even if it fails). Is that what we want?

resources/assets/js/store/Application/applicationSelector.ts
```diff
@@ -39,6 +39,23 @@ const constructNonNormalizedApplication = (
...
+
+export const getApplication = createCachedSelector(
+  getApplicationState,
+  (state: RootState, ownProps: { applicationId: number }): number =>
+    ownProps.applicationId,
+  (applicationState, applicationId): ApplicationNormalized | null =>
+    hasKey(applicationState, applicationId)
+      ? applicationState[applicationId]
+      : null,
+)((state, ownProps): number => ownProps.applicationId);
```
tristan-orourke, line 57: I'm not sure this selector gains anything from using (re-)reselect. As far as I know, the main reason to cache selectors (besides when you have a computationally expensive selector) is to avoid returning something like a new array object (which will cause re-renders) when the contents are the same.

resources/assets/js/components/Application/BasicInfo/BasicInfoPage.tsx
Khristophor: Ideally yeah, we should get some loading UI from design

tristan-orourke:
Oh, and a couple tests are failing because they're still using 'post' instead of 'put' for the route that was changed.

tristan-orourke (approved):

PR #4395, opened 2020-09-17
Update the CheckboxGroup component

tristan-orourke:
Replaces the CheckboxGroup component with a new simpler version, CheckboxGroupField.

Advantages over old component: t
- his manages the whole Formik checkbox group within a single component
- it doesn't require passing in any Formik properties
- it doesn't set the id attribute of any elements (previous version set id to match name, which caused duplicate ids when multiple similar forms were on one page)


Khristophor (commented):
Looks good, just need to fix up the Testcafe test that's failing 👍

yonikid15 (changes requested):
There's one issue. It's not focusing on the checkbox group field if it's invalid.

tristan-orourke:
The tests are fixed now.

yonikid15 (approved):

Khristophor (approved):

PR #4416, opened 2020-09-22
Reuse Application data fetching code between steps

tristan-orourke:
### Notes:
- Each step now triggers requests to fetch any un-loaded data which is required for any step, not just the current step. This will prevent further loading states when a user reaches a later step.
- Each piece (or set) of data has a hook which determines whether it is present in the redux store, and triggers a fetch if it is not. There is a single hook which combines all of these, which is used in each step.
- The redux selectors have also been extracted to their own hooks for simplicity.


Khristophor (approved):
Nice work! This is pretty much what I had pictured for the application, but this covers everything. My only suggestion would be to move the hooks to the hooks folder we already have.

tristan-orourke:
> Nice work! This is pretty much what I had pictured for the application, but this covers everything. My only suggestion would be to move the hooks to the hooks folder we already have.

Thanks! I'll move applicationHooks to the hooks folder. And now that the Fit step is merged into dev, I'll refactor that step as well.

PR #4421, opened 2020-09-23
Task/allow indeterminate job posters

### LINKED ISSUE #4226: Task - Admin - Add option for departments to "Allow indeterminate job posters"
# Description
Add a checkbox to the department on Admin panel for "Allow indeterminate job posters", which will allow posters built for this department to avoid setting a term length.

This will require the following steps:
1. Write a database migration which adds a boolean field to the 'departments' table.
2. Add the new field to the documentation and $fillable array of the Department model.
3. Add a field to the DepartmentCrudController, which controls the Department page of the Admin portal (which is built with Laravel Backpack).

jaysonmc:
Resolves #4226

### Notes: 
- Not sure if there is usually any testing that accompanies changes like this. Quick search in the tests folder makes me think perhaps not. But if Im mistaken and I missed the tests let me know and they shall be added.

### DIFF
--- a/app/Http/Controllers/Admin/DepartmentCrudController.php
+++ b/app/Http/Controllers/Admin/DepartmentCrudController.php
@@ -51,7 +51,7 @@ class DepartmentCrudController extends CrudController
             $this->crud->addField([
                 'name' => 'indeterminate',
                 'type' => 'checkbox',
-                'label' => 'indeterminate',
+                'label' => 'Indeterminate',
             ]);
         });
     }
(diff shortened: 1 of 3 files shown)


tristan-orourke (changes requested):
Everything works correctly, I would just rename the field name to 'allow_indeterminate', to be a little more clear about its purpose.

tristan-orourke:
And yes, for something like this we wouldn't normally write tests.

tristan-orourke (approved):

PR #4455, opened 2020-10-01
Copy course / degree / work experience data to new models

### LINKED ISSUE #4409: Task - Copy existing Course, Degree & Experience objects to new ExperienceEducation and ExperienceWork objects
# Description
Courses, Degrees, and Equivalent Experience map 1-to-1 to ExperienceEducation and ExperienceWork. We must write migrations that create new Experience objects for each of these old objects, with the data copied over.

gggrant:
Resolves #4409.

Only copies data, no changes to the existing tables.

### DIFF
--- a/database/migrations/2020_10_01_180409_copy_degrees_and_courses_to_experiences_education.php
+++ b/database/migrations/2020_10_01_180409_copy_degrees_and_courses_to_experiences_education.php
@@ -16,6 +16,7 @@ class CopyDegreesAndCoursesToExperiencesEducation extends Migration
         $degrees = DB::table('degrees')->get();
 
         foreach ($degrees as $degree) {
+            $status_id = DB::table('education_statuses')->where('key', 'complete_credited')->first()->id;
 
             if ($degree->blockcert_url) {
                 $blockcert = true;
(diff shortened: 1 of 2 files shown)


database/migrations/2020_10_01_180409_copy_degrees_and_courses_to_experiences_education.php
```diff
@@ -0,0 +1,70 @@
...
+                $blockcert = true;
+            } else {
+                $blockcert = false;
+            }
+
+            DB::table('experiences_education')->insert([
+                'education_type_id' => $degree->degree_type_id,
+                'area_of_study' => $degree->area_of_study,
+                'institution' => $degree->institution,
+                'education_status_id' => 1,
```
tristan-orourke, line 30: I'd prefer to get the id with a query, something like `$complete_id = DB::table('education_statuses)->where('name', 'complete_credited')->first()->id` (that's probably the wrong syntax). Just makes it a bit more robust and readable.

tristan-orourke (changes requested):
Lets be explicit on the meanings of the foreign ids, but otherwise it looks good. I'm also waiting on confirmation from Gray about copying WorkExperiences to ExperiencesWork.

database/migrations/2020_10_01_180409_copy_degrees_and_courses_to_experiences_education.php
gggrant: You got it.

tristan-orourke (approved):
All good. Just important to remember that we can't push this live until we're ready to launch the new Application.

PR #4462, opened 2020-10-02
Manager/HR - 'Return to Summary' button on review / preview

### LINKED ISSUE #2304: Task - Manager/HR - Add Back to Summary button on Job Poster Review and Preview
# Description
### For reference:
- The **Review** page is the last page of the Job Poster Builder
- The **Preview** page is the /job/# page that will eventually become the "live" job poster page
- The **Summary** page is the new summary page for a single job poster that we're building as part of the HR portal epic
- The **Job List** is the page where all of a manager's or HR advisers jobs are listed. This was previously referred to as the dashboard.

## Task
At the bottom of the Review and Preview pages, add a button that navigates back to the Summary page. The button should simply say 
EN - "Save and Return to Summary"
FR - "Enregistrer et retourner au sommaire"

This would replace the "looks good" button on the review.

gggrant:
Resolves #2304.


yonikid15 (commented):
The button's text needs to also change depending on the state of the job poster. Currently, the **Save and Return to Preview** button is always validating and submitting the poster, which is causing an error if the poster is live. A solution would be to do the following:
- If the job poster status is `draft`, then display **Submit**.
- If the job poster has been submitted and going through the editing/approval process, then display **Save and Return to Preview**.
- If the job poster is `live` (or any status past live), then just use **Return to Summary** link. Also, we should remove all edit links, and **Save & Return to Skills** button.

@gobyrne What do you think?

gobyrne:
@yonikid15 actually, the point is that we don't want this button to submit the job poster at all anymore. The only place you should be able to submit is on the Summary page. This button should just run the validation and save the data.

yonikid15:
Oh okay. Then well have to remove or change the text on the congrats modal that pops up after saving and returning to summary page. 

image

resources/assets/js/components/JobBuilder/Review/JobReview.tsx
```diff
@@ -860,7 +861,7 @@ export const JobReview: React.FunctionComponent<JobReviewProps &
             >
               <FormattedMessage
                 id="jobBuilder.review.button.submit"
-                defaultMessage="Looks good!"
+                defaultMessage="Save and Return to Summary"
                 description="Label of Job Review Submission Button"
```
yonikid15, line 865: Last thing to do is changing this button into an link (which can look like a button). Also, add a comment saying the Congrats modal is will be moved to the summary page.

yonikid15: @gggrant I already have it open so I can just pop it in for you.

gggrant: Awesome, thanks!

yonikid15 (approved):

PR #4477, opened 2020-10-06
Update local setup instructions

mnigh:
### Notes:
- README for new members of the team to install from scratch using macOS

### DIFF
--- a/README-local.md
+++ b/README-local.md
@@ -315,33 +315,23 @@ These instructions were written using macOS High Sierra (Version 10.13.6)
 /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install.sh)"
 ```
 
-## Install nginx
+## Install composer
 
 ```
-brew install nginx
-brew services start nginx
-sudo ufw allow 'Nginx HTTP'
+sudo brew install composer
 ```
 
-## Install PostGres
+## Install laravel/valet
+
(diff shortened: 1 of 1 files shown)


README-local.md
```diff
@@ -287,6 +291,123 @@ extension=imagick
...
+
+#### Setup TalentCloud directory to be hosted
+
+*For more details, please visit [here](https://laravel.com/docs/8.x/valet#securing-sites)*
+
+```
+cd TalentCloud
+valet secure talent
+valet link talent
+valet restart
```
Khristophor, line 388: Curious why you're installing nginx and running valet? AFAIK valet should take care of an nginx installation and running it in the background when the Mac starts. Looks like you can also tell valet to use a specific version of PHP via Homebrew according to [the docs](https://laravel.com/docs/7.x/valet#installation): `valet use php@7.2`.

Not opposed to one method or another, but it might make sense to outline these as alternatives rather than part of the same setup process.

Khristophor (changes requested):

README-local.md
mnigh: Thanks for the feedback @Khristophor. I have updated the instructions for the macOS portion to remove the part to install nginx via homebrew, added `valet use php@7.2` into the process once valet has been installed, and cleaned up a few other sections to hopefully make the process more straightforward.

Khristophor (approved):

PR #4508, opened 2020-10-14
Add Application Timeline validation

Khristophor:
### Notes:
- New Timeline validation class with unit testing
- Need to add translation lines for a new custom rule and simplified step messaging:
```
Missing required :relation equal to :attributeValue

Step [x] must be completed
```


gobyrne:
> Missing required :relation equal to :attributeValue

Il manque la valeur obligatoire pour :relation égal à :attributeValue


> Step [x] must be completed

L'étape [x] doit être complétée


**These should never be exposed to end-users, so we're not worried about their clarity**

yonikid15 (changes requested):
Found an issue, and not sure why yet. It looks like the experiences step is still getting validated as true when there are zero experiences assigned to a essential criteria. Same issue is happening on the Skill step. If you go to the `setUp()` method in the test file, and remove the section at the end (starts at line 90), which creates the ExperienceSkills, `testExperienceComplete()` still returns true.

yonikid15 (changes requested):

app/Services/Validation/ApplicationTimelineValidator.php
```diff
@@ -0,0 +1,421 @@
...
+                    ['experienceable_id', $application->id]
+                ]);
+            }
+        )
+        ->whereIn('skill_id', $requiredCriteria->pluck('skill_id')->all())
+        ->get();
+
+        $validator = Validator::make($experiences->toArray(), $experienceSkillRules);
+
+        return $validator;
```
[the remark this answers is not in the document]
Khristophor: So this bit:
```php
->whereIn('skill_id', $requiredCriteria->pluck('skill_id')->all())
```
should be limiting the query to only bring back relevant experiences (if they exist). We can definitely add an orderBy to ensure they both line up, if that helps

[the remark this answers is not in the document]
Khristophor: Maybe the validator itself is the issue, I had assumed the `ContainsArrayWithAttributeRule` would validate as long as the `$experiences` array had one of each of a `skill_id` of a given number. If it's not working that way, then I guess I need to adjust the validation logic.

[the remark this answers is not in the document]
yonikid15: Ah okay cool good to know. However, ya we'll need to refactor it cause it, like I said earlier it still passes when `$experiences` are empty. Actually, as long as the `ContainsArrayWithAttributeRule` rules never lines up with an incorrect `skill_id` then it passes. Even if the remaining rules aren't met.

app/Services/Validation/ApplicationTimelineValidator.php
```diff
@@ -0,0 +1,420 @@
...
+            return $criterion->criteria_type_id === $essentialCriteriaType;
+        });
+
+        $experienceSkillRules = array();
+
+        foreach ($requiredCriteria as $criterion) {
+            // Validate that every essential skill has a corresponding experience.
+            $experienceSkillRules[] = [
+                'skill_id' => new ContainsArrayWithAttributeRule('skill_id', $criterion->skill_id)
+            ];
```
tristan-orourke, line 120: So experienceSkillRules will end up looking like:
```
[
  ['skill_id' => ContainsArrayWithAttributeRule('skill_id', 1)],
  ['skill_id' => ContainsArrayWithAttributeRule('skill_id', 2)],
]
```
Is this right? Do we need the nested arrays? 

Also, should each rule have the 'skill_id' key? If I understand the rule, this will check that the 'skill_id' field of the validated object is an array that _also_ has a 'skill_id' field.

app/Services/Validation/ApplicationTimelineValidator.php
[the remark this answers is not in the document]
tristan-orourke: I think the ContainsArrayWithAttributeRule doesn't work as you expect. The way I read it, this will check that the 'skill_id' field of the validated object is an array that also has a 'skill_id' field (with the correct value). I'm also not sure how this will work when the validated object is itself an array.

app/Services/Validation/ApplicationTimelineValidator.php
```diff
@@ -0,0 +1,420 @@
...
+     */
+    public function experienceValidator(JobApplication $application)
+    {
+        $application->loadMissing('job_poster', 'job_poster.criteria');
+
+        $essentialCriteriaType = CriteriaType::where('name', 'essential')->first()->id;
+
+        $jobCriteria = $application->job_poster->criteria;
+        $requiredCriteria = $jobCriteria->filter(function ($criterion) use ($essentialCriteriaType) {
+            return $criterion->criteria_type_id === $essentialCriteriaType;
```
tristan-orourke, line 111: We need to make sure we're only requiring Hard Skills as well.

app/Services/Validation/ApplicationTimelineValidator.php
```diff
@@ -0,0 +1,420 @@
...
+     */
+    public function skillsValidator(JobApplication $application)
+    {
+        $application->loadMissing('job_poster', 'job_poster.criteria');
+
+        $essentialCriteriaType = CriteriaType::where('name', 'essential')->first()->id;
+
+        $jobCriteria = $application->job_poster->criteria;
+        $requiredCriteria = $jobCriteria->filter(function ($criterion) use ($essentialCriteriaType) {
+            return $criterion->criteria_type_id === $essentialCriteriaType;
```
tristan-orourke, line 174: Check its for a Hard skill as well.

app/Services/Validation/ApplicationTimelineValidator.php
```diff
@@ -0,0 +1,420 @@
...
+            ];
+        }
+
+        // Get all Experiences belonging to the current application, that are assigned to required Criteria.
+        $experiences = ExperienceSkill::whereHasMorph(
+            'experience',
+            '*',
+            function (Builder $query) use ($application): void {
+                $query->where([
+                    ['experienceable_type', 'application'],
```
tristan-orourke, line 129: The experienceable_type depends on the $application status. If `$application->isDraft()`, experienceable_type must be 'applicant'. (And since we mainly need to validate _before_ submitting, this will usually be the case.)

app/Services/Validation/ApplicationTimelineValidator.php
```diff
@@ -0,0 +1,420 @@
...
+
+        $essentialCriteriaType = CriteriaType::where('name', 'essential')->first()->id;
+
+        $jobCriteria = $application->job_poster->criteria;
+        $requiredCriteria = $jobCriteria->filter(function ($criterion) use ($essentialCriteriaType) {
+            return $criterion->criteria_type_id === $essentialCriteriaType;
+        });
+
+        // Validate that every experience has a justification.
+        $experienceSkillRules = array('*.justification' => 'required|max:100');
```
tristan-orourke, line 178: We can also specify its a string in the rule.

app/Services/Validation/ApplicationTimelineValidator.php
tristan-orourke: And of course, it type is 'applicant', id will need to be `$application->applicant_id`.

tests/Unit/Validators/ApplicationTimelineValidatorTest.php
```diff
@@ -0,0 +1,276 @@
...
+        foreach ($this->experienceSkills as $experience) {
+            $experience->justification = $this->faker->text(99);
+            $experience->save();
+        }
+        // Every essential criterion has an assigned experience above, this should
+        // validate.
+        $this->assertTrue($validator->skillsComplete($this->application));
+
+        // An empty justification should fail validation.
+        $this->experienceSkills[0]->justification = null;
```
tristan-orourke, line 210: Maybe try this with an empty string as well?

tristan-orourke (changes requested):
The tests look nice and thorough! I'm quite confused about how the Experience validation is working, but I'll work with Yoni to understand and help fix it.

tristan-orourke:
This isn't particularly tied to this PR, but it just occurred to me that if we have a separate lookup table for Application versions, we should add some kind of title or description column.

app/Services/Validation/ApplicationTimelineValidator.php
```diff
@@ -0,0 +1,429 @@
...
+     */
+    public function experienceValidator(JobApplication $application)
+    {
+        $application->loadMissing('job_poster', 'job_poster.criteria');
+
+        $essentialCriteriaType = CriteriaType::where('name', 'essential')->first()->id;
+        $hardSkillType = SkillType::where('name', 'hard')->first()->id;
+
+        $jobCriteria = $application->job_poster->criteria;
+        $requiredCriteriaSkillIds = $jobCriteria
```
tristan-orourke, line 114: I renamed requiredCriteria to requiredCriteriaSkillIds, and a couple other similar changes, to better reflect what they are now.

tests/Unit/Validators/ApplicationTimelineValidatorTest.php
```diff
@@ -0,0 +1,253 @@
...
+        $validator = new ApplicationTimelineValidator();
+        // Draft Application
+        $draftApplication = factory(JobApplication::class)->states(['draft', 'version2'])->create();
+        // Every essential criterion has an assigned experience above, this should
+        // validate.
+        $this->assertTrue($validator->experienceComplete($draftApplication));
+        // Removing one or more experiences from a required criterion should fail
+        // validation.
+        $essentialCriteriaType = CriteriaType::where('name', 'essential')->first()->id;
+        $requiredSkill = $draftApplication->job_poster->criteria
```
tristan-orourke, line 127: When looking for a required skill (here and on other tests in the file) make sure to check that its a 'hard' skill as well. As is, it will occasionally fail.

tristan-orourke (changes requested):
Looks good now, except for a couple tests that can fail sporadically.

tristan-orourke (approved):

yonikid15 (approved):
👍 👍

PR #4509, opened 2020-10-15
Timeline - Application step statuses 

### LINKED ISSUE #2870: Task - Application - Determine completion state of each Application step on the frontend

yonikid15:
Resolves #2870 

The validation PR #4508 must be merged before this.

### Notes:
- Applicants can reach the next step of the application, only if they have completed the previous step. Therefore, I created a new `appliction_steps` table which holds the application step as well as a `touched` field. If the step has been set to `touched`, we absolutely know that the previous step was completed at some point. 
- This is a draft PR since it still requires a couple changes/fixes:
    - [x] Adding validation for each step from Chris's branch.
    - [x] Preventing applicants from jumping ahead to step by manually entering the url.


app/Models/JobApplication.php
```diff
@@ -45,7 +48,7 @@
  * @property string $gov_email
  * @property boolean $physical_office_willing
  * @property int $security_clearance_id
- * @property boolean $share_with_maangers
+ * @property boolean $share_with_mangers
```
tristan-orourke, line 51: Oops, you fixed a typo, but its still a typo :P

resources/assets/js/components/Application/RedirectToLastTouchedStep.tsx
```diff
@@ -0,0 +1,119 @@
...
+import SkillsIntroPage from "./Skills/SkillsIntroPage";
+import SkillsPage from "./Skills/SkillsPage";
+
+interface RedirectToLastTouchedStepProps {
+  applicationId: number;
+  requestedStep?: ApplicationStep;
+  introStep?: boolean;
+}
+
+const RedirectToLastTouchedStep: React.FunctionComponent<RedirectToLastTouchedStepProps> = ({
```
tristan-orourke, line 29: I have a few different thoughts about this component. 

My first thought was it would seem more appropriate to move this logic to the Laravel controller, do a redirect if necessary before the request is returned, and leave the ApplicationRoot router component as it was. This might be faster, since the redirect could be determined before the view is rendered.

If we do want to keep this Redirect component, then instead of just rendering the correct Step component, it should call the  redirect function in helpers/router.tsx, so that the url matches the step shown.

My last thought is that this all seems like overkill. I can understand not showing links in the Progress Bar to un-touched steps, but if someone goes through all the trouble of typing out the url, why stop them? We've built each step not to break if visited out of order.

resources/assets/js/store/Application/applicationActions.ts
```diff
@@ -237,6 +245,40 @@ export const sendReferenceEmail = (
...
+
+export type UpdateApplicationStepAction = AsyncFsaActions<
+  typeof UPDATE_APPLICATION_STEP_STARTED,
+  typeof UPDATE_APPLICATION_STEP_SUCCEEDED,
+  typeof UPDATE_APPLICATION_STEP_FAILED,
+  { [step in ApplicationStep]: ProgressBarStatus },
+  { applicationId: number; stepId: number }
+>;
+
+export const updateApplicationStep = (
```
tristan-orourke, line 262: Can we rename this 'touchApplicationStep', or add a comment explaining that it just sets 'touched' to true, regardless of the Body of the PUT request?

resources/assets/js/store/Application/applicationActions.ts
```diff
@@ -237,6 +245,40 @@ export const sendReferenceEmail = (
...
+): RSAActionTemplate<
+  typeof UPDATE_APPLICATION_STEP_STARTED,
+  typeof UPDATE_APPLICATION_STEP_SUCCEEDED,
+  typeof UPDATE_APPLICATION_STEP_FAILED,
+  { [step in ApplicationStep]: ProgressBarStatus },
+  { applicationId: number; stepId: number }
+> =>
+  asyncPut(
+    getUpdateApplicationStepEndpoint(applicationId, stepId),
+    { applicationId, stepId },
```
tristan-orourke, line 274: The body of the request should be null, since its not used on the backend.

tests/Feature/Api/ApplicationControllerTest.php
```diff
@@ -110,4 +111,28 @@ public function testUpdateBasicsValidation(): void
...
+
+
+        $data = [
+            'applicationId' => $application->id,
+            'stepId' => $step->id,
+        ];
+
+        $response = $this->actingAs($applicant->user)
+        ->json('put', "$this->baseUrl/applications/$application->id/steps/$step->id", $data);
+        $response->assertOk();
```
tristan-orourke, line 133: This request doesn't need any data.

tests/Feature/Api/ApplicationControllerTest.php
```diff
@@ -110,4 +111,28 @@ public function testUpdateBasicsValidation(): void
...
+
+        $application->refresh();
+        $response->assertJsonFragment($application->applicationTimelineSteps());
```
tristan-orourke, line 136: Suggestions for more test cases:
- Assert that the 'basic' step is touched, while the others are not. The current assertion would pass if the request returned the steps without changing any.
- Test that calling the request on the same step multiple times doesn't do anything weird.

tristan-orourke (changes requested):

database/migrations/2020_10_09_143603_create_steps_table.php
```diff
@@ -0,0 +1,32 @@
+<?php
+
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+
+class CreateStepsTable extends Migration
```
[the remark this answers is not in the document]
yonikid15: I was thinking to make this Step model reusable for other features, if need be. However, that might be a bit overkill actually.   Also, it would need another column to determine which feature it belongs to (eg. `feature`, `type`).

resources/assets/js/components/Application/BasicInfo/BasicInfoPage.tsx
```diff
@@ -55,6 +61,9 @@ const BasicInfoPage: React.FunctionComponent<BasicInfoPageProps> = ({
     values: ApplicationNormalized,
   ): Promise<void> => {
     await updateApplication(values);
+    await dispatch(
+      updateApplicationStep(applicationId, ApplicationStepId.experience),
+    );
```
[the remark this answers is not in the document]
yonikid15: I think it's better to update the state here, because if we reach the next step before the request is done then you'll get a flash  of the error status on the ProgressBar. We could avoid this by checking if the steps are loading and adding a new `loading` state, with the loading icon. However, I think it's cleaner to leave it like this.

app/Models/JobApplication.php
```diff
@@ -11,10 +11,13 @@
 use App\Events\ApplicationSaved;
 use App\Models\Applicant;
 use App\Models\ApplicationReview;
+use App\Models\Lookup\JobApplicationStep;
+use App\Services\Validation\ApplicationTimelineValidator;
 use App\Services\Validation\ApplicationValidator;
 use App\Services\Validation\StrategicResponseApplicationValidator;
 use Illuminate\Notifications\Notifiable;
 use App\Traits\TalentCloudCrudTrait as CrudTrait;
+use Illuminate\Support\Facades\Log;
```
tristan-orourke, line 20: Log can be deleted

resources/assets/js/components/Application/Intro/IntroPage.tsx
```diff
@@ -1,11 +1,10 @@
 /* eslint-disable camelcase */
-import React from "react";
+import React, { useState } from "react";
```
tristan-orourke, line 2: useState isn't used

tristan-orourke (changes requested):
I'm not yet convinced that the Application Step components are sending the touch requests in the right place - I'll branch off and try a few things first. In the meantime, there's just a few other minor changes that could be made.

app/Http/Controllers/ApplicationTimelineController.php
```diff
@@ -14,8 +16,46 @@ class ApplicationTimelineController extends Controller
...
+        // in the application then redirect back to last touched step.
+        $reversedStepOrder = array_reverse($stepOrder);
+        foreach ($reversedStepOrder as $lastTouchedStep) {
+            if ($jobApplicationSteps[$lastTouchedStep] === 'complete' || $jobApplicationSteps[$lastTouchedStep] === 'error') {
+                return redirect()->route('application.timeline', ['jobApplication' => $jobApplication, 'step' => $lastTouchedStep]);
+            }
+        }
+
+        // If no step has been reached yet it will take applicant to welcome step.
         return view('applicant/application-timeline-root')
```
tristan-orourke, line 59: This won't necessarily take you to the welcome step - it'll take you to whichever step you wanted to go to. That means that you _can_ skip ahead to a future step, if you haven't clicked through any of the application yet.
This suggest this show() function contains enough logic that it should have its own unit tests.

yonikid15: Hey, so I added in the tests. I had to make some new routes for the application timeline in the applicant group, since the demo routes don't work. After you check out the tests, we can just comment out everything and make a note to uncomment later. We can also just change the paths of the route to something different then the old applications route, I believe.

yonikid15:
Hey @tristan-orourke, so I added in the tests. I had to make some new routes for the application timeline in the applicant group, since the demo routes don't work. After you check out the tests, we can just comment out everything and make a note to uncomment later. We can also just change the paths of the route to something different then the old applications route, I believe.

app/Models/TouchedApplicationStep.php
```diff
@@ -0,0 +1,44 @@
...
+{
+    protected $casts = [
+      'job_application_id' => 'int',
+      'step_id' => 'int',
+      'touched' => 'boolean',
+    ];
+
+    protected $fillable = [
+      'job_application_id',
+      'step_id',
```
tristan-orourke, line 29: I don't these two should be fillable. They should be immutable after the row is created.

resources/assets/js/components/Application/BasicInfo/BasicInfoPage.tsx
```diff
@@ -54,7 +63,7 @@ const BasicInfoPage: React.FunctionComponent<BasicInfoPageProps> = ({
   const handleContinue = async (
     values: ApplicationNormalized,
   ): Promise<void> => {
-    await updateApplication(values);
+    Promise.all([await updateApplication(values)]);
```
tristan-orourke, line 66: Promise.all is unnecessary now

resources/assets/js/components/Application/Intro/IntroPage.tsx
```diff
@@ -44,13 +48,25 @@ export const IntroPage: React.FunctionComponent<IntroPageProps> = ({
...
+        <h2
+          data-c-heading="h2"
+          data-c-align="center"
+          data-c-padding="top(2) bottom(3)"
+        >
+          {intl.formatMessage(loadingMessages.loading)}
+        </h2>
+      )}
+      {application !== null && job !== null && (
+        <Intro handleStart={handleContinue} />
```
tristan-orourke, line 68: The intro page doesn't actually need the application or job, so no need to wait for them to load.

resources/assets/js/components/Application/Intro/IntroPage.tsx
```diff
@@ -44,12 +48,22 @@ export const IntroPage: React.FunctionComponent<IntroPageProps> = ({
...
       )}
+      {showLoadingState && (
+        <h2
+          data-c-heading="h2"
+          data-c-align="center"
+          data-c-padding="top(2) bottom(3)"
+        >
+          {intl.formatMessage(loadingMessages.loading)}
+        </h2>
+      )}
```
tristan-orourke, line 66: I don't think we should show this loading state at all, since we're not waiting to show anything.

tristan-orourke (approved):

PR #4510, opened 2020-10-15
Enable route caching

Khristophor:
### Notes:
- To test, use the command `php artisan route:trans:cache` to cache routes, and `php artisan route:clear` to clear caching
- Locally I see a ~70 - 80 ms reduction in boot time
Before:
Screen Shot 2020-10-15 at 12 36 32 PM
After:
Screen Shot 2020-10-15 at 12 37 00 PM

### DIFF
--- a/routes/web.php
+++ b/routes/web.php
@@ -120,14 +120,10 @@ Route::group(
 
                 /* Response Home */
                 // Redirect /en/response to /response so it reaches the Talent Reserve app.
-                Route::get('response', function () {
-                    return redirect(URL::to('/response'));
-                });
+                Route::redirect('response', URL::to('/response'));
 
                 /* Reserve Redirect */
-                Route::get('reserve', function () {
-                    return redirect('response');
-                });
+                Route::redirect('reserve', URL::to('/response'));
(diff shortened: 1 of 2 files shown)


routes/web.php
```diff
@@ -287,37 +283,29 @@ function (): void {
...
-                                    Lang::get('common/privacy.title') => '',
-                                ],
-                            ]
-                        );
-                    }
+                    'common/static_privacy',
+                    [
+                        'privacy' => Lang::get('common/privacy'),
+                        'custom_breadcrumbs' => [
+                            'home' => LaravelLocalization::localizeUrl('/'),
```
Khristophor, line 292: Composer/PHP throws an error here when you try and use the `route()` helper to reference a named route. I assume it's because it's not actually defined yet when it tries to cache the entire routes file.

tristan-orourke (approved):
Looks good. Oddly, I don't see an improvement after running the route cache command on my local machine, but I trust it can't hurt!
If we're making caching possible, we might as well trigger it in the deploy script, so I made that change.

PR #4556, opened 2020-10-22
Bugfix - Intermittent test failures around Skill Declaration description validation

Khristophor:
### Notes:
- Previously using `$faker->paragraphs(3, true);` would occasionally generate more than the 100 character max validation limit, causing a test failure
- Unsure if this is responsible for all the failures relating to the skillsComplete step, but I ran this test consecutively for over an hour locally with no failures


yonikid15 (approved):
👍

PR #4575, opened 2020-10-26
Application Changes Static Info Page

JoshBeveridge:
Adds a static info page in EN/FR at /application-update.

Educates users on the upcoming change to the application process.

Waiting on more info as to when this will be going live.


gobyrne:
@JoshBeveridge should we close this now? Doesn't seem relevant

Issue #4548, opened 2020-10-21
Bug - Manager - Job Builder skill level text overflow


mnigh:
# Description
When skill level text is longer than the column in which it sits, the pill element breaks to a new line.

# Technical Info
**Operating System:** macOS
**Browser:** Chrome

# Steps to Reproduce
1. Go to Step 5 of Job Builder in French (Skills/Compétences)
2. Click on **Orientation de service**
3. Select _Démonstration à un niveau profond_ or _Phase de développement précoce_
4. See text under **Compétences essentielles** heading.

# Images
![skill-level-long](https://user-images.githubusercontent.com/3046459/96787974-ac484c80-13c0-11eb-9d70-dd3358742cb0.png)

PR #4577, opened 2020-10-26
Fixes bug #4548 by making the level spans inline-block.

bitliege:
New PR for merging into the correct release.

Text wraps inside spans now.


Khristophor:
Reopened as #4590 to ensure the build won't break when merged

PR #4581, opened 2020-10-26
Fully functional Profile Experience page

### LINKED ISSUE #4084: Applicant - Connect new Profile page to data source
# Description
The new profile will need to load (from the Redux store) all the Experiences and ExperienceSkills associated with the Applicant.

tristan-orourke:
Resolves #4084.

Fully replaced the existing Profile Experiences page with the new one based on the new Application.
See it at https://talent.test/en/profile/{applicantId}/experience

### Notes:
- As I realized there were more and more differences between the experience page on the Application and on the Profile, I split them into two different components, instead of one component used for both.
- Deleting experience or experienceSkills is causing React warnings in the console -  probably has to do with promises in accordion components that return after the accordion is unmounted.

React warning from console


Khristophor (changes requested):
This is going to be a pain, but I'd recommend adding the [readonly HTML attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/readonly) to the various fields that support it to the archived sections of the profile.

It will allow users to select and copy text out of input fields, but prevent them from adding new content (or accidentally replacing existing content).

Khristophor (changes requested):
I know it's in the designs, but it feels weird to have disabled buttons next to all of these sections that users can no longer interact with.

I can also just remove the `disabled` attribute, and continue on as normal:
Screen Shot 2020-10-27 at 11 06 12 AM
Screen Shot 2020-10-27 at 11 06 42 AM

**Edit:**
I guess we aren't looking to put any effort into reworking the UI for the archived sections, but we should still having some client/server side checks so users can't add information from these views.

bitliege:
In this case, the designs should have presented read only versions of each form UI that can be presented to the user. @Khristophor is correct that it's confusing to include the buttons, but it's worse to continue presenting the information in a form format as well.

tristan-orourke:
@Khristophor @bitliege You're right that its easy to remove the disabled tag and submit the forms like before. I think we should remove the server endpoints used by these forms to update data, but not worry too much about the UX on these archived pages. I've created a separate task for removing the endpoints, since it doesn't really depend on this PR: #4605.

Khristophor (approved):

PR #4590, opened 2020-10-27
Bugfix - French text wrapping on Criteria name

Khristophor:


Khristophor:
I guess our Travis config won't trigger builds for this PR since it's merging into a release branch, but I think it's safe to merge in

yonikid15 (commented):
There's a clone attribute that you can use: `data-c-display="inline-block"`.
Also, I think it looks better centered `data-c-alignment="center"`, but that's just me.

yonikid15 (approved):
✔️

PR #4594, opened 2020-10-27
Touch application step when it loads, instead of before navigating to it

tristan-orourke:
I propose this alternate method of touching application steps. This way, navigating to the next step doesn't have to wait on this request finishing.
It's also more robust. The method of touching a step before navigating to it will cause bugs if we later allow the steps to be visited in any other order.


resources/assets/js/components/Application/BasicInfo/BasicInfoPage.tsx
```diff
@@ -46,6 +43,8 @@ const BasicInfoPage: React.FunctionComponent<BasicInfoPageProps> = ({
   const job = useJob(jobId);
   const steps = useJobApplicationSteps();
 
+  useTouchApplicationStep(applicationId, "basic", dispatch);
```
yonikid15, line 46: I'm getting a flash of the error state. When the user hits that step for the first time, on the backend it sets the step to touched. Then the response payload has that step set to `error`. When they complete the step and move to the next, that's when you'll see a flash of the error state.  

To prevent this we can add a loading state. This would mean adding a `stepIsUpdating` state to the applications reducer, adding new action, etc.

yonikid15 (changes requested):

resources/assets/js/components/Application/BasicInfo/BasicInfoPage.tsx
tristan-orourke: I see what you mean now.

I think adding a stepIsUpdating state is the right approach. In terms of what to show for loading state - maybe just replace any 'error' states with 'default' temporarily, as 'error' is the most disruptive.

tristan-orourke: Okay, I've made the changes I just mentioned.

Khristophor:
Not at all relevant to this PR, but this plays in my head whenever we talk about touching application steps

https://www.youtube.com/watch?v=A52--FKUQgU

yonikid15 (approved):

PR #4620, opened 2020-10-30
Task/job get classification

### LINKED ISSUE #3908: Task - Manager - Pull classification from database for JPB
# Description
The job poster builder has a list of classifications built into the front-end, but we also have a list of classifications in the database that are used on the backend. The frontend should be retrieving the list of from the backend to avoid mismatches. This will also allow Admins to edit the list of classifications, and expect the results to appear in the app.

To complete this task:
1. Create an api endpoint that returns all classifications. This means adding a route in the api section of routes/web.php, and creating a new Controller.
2. Add Classifications to the frontend Redux store. This involves an unfortunate amount of boilerplate, as we need to add actions, selectors, and reducer.
3. Delete ClassificationId from lookupConstants.ts, and classificationCodes in localizedConstants.tsx. Ensure anywhere that was using these values, instead requests them and then selects them from the redux store.

jaysonmc:
Resolves #3908

### Notes:
The issue says "Delete ClassificationId from lookupConstants.ts, and classificationCodes in localizedConstants.tsx."

Doing this will break all the pages that use this, and I'm not entirely sure where all these values are used in the application. Given the number of files changed already after a discussion with @tristan-orourke, it was decided to merge the current state to avoid the likelihood of a nightmare merge

Option 1: We can either leave the branch open and add the missing work (delete the values in the config and fix all the files that break)
Option 2: Merge this PR and create a new issue to plug in the values via redux (instead of the config files)


database/migrations/2020_10_19_151339_add_desc_and_defaultmsg_to_classification.php
```diff
@@ -0,0 +1,88 @@
...
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('classifications', function (Blueprint $table) {
+            $table->string('defaultMessage')->nullable();;
+
+            $table->string('description')->nullable();
```
Khristophor, line 19: I'm assuming we're returning both of these to the frontend, so they'll need to be localized. We're using the [Laravel translatable package](https://github.com/spatie/laravel-translatable) for localizing our model fields, so you'll need to make these columns json instead of string.

This is an example of where we added this type of data before: https://github.com/GCTC-NTGC/TalentCloud/blob/dev/database/migrations/2020_02_24_083132_add_name_and_description_to_job_poster_status.php

resources/assets/js/models/types.ts
```diff
@@ -98,7 +98,10 @@ export interface AwardRecognitionType {
 }
 
 export interface Classification {
+  id: number;
   key: string;
+  defaultMessage: string;
+  description: string;
```
Khristophor, line 104: After implementing the migration changes, these will need to be typed as `localizedField[Non]Null` to make use of the expected json structure

routes/web.php
```diff
@@ -1080,4 +1080,6 @@ function (): void {
         ->where('application', '[0-9]+')
         ->middleware('can:view,application')
         ->name('application.experience.index');
+    Route::get('classifications', 'ClassificationsController@getClassifications');
+        //->middleware('can:view,application');
```
Khristophor, line 1084: I don't think we necessarily need middleware here, as most of our other lookup data is open. I'm also thinking we might want to leave this under the `v1` group and put it around line 858 just for consistency with the other lookup endpoints
```php
Route::get('classifications', 'Api\ClassificationController@index');
```

Khristophor (changes requested):

jaysonmc:
@Khristophor all comments should be addressed. Let me know if I missed anything.

Khristophor (approved):
Looking good for now, we'll need to spin off another story/PR to address the `ClassificationID` constant that needs removing (it's a terrible rabbit hole, I attempted it this morning)

PR #4622, opened 2020-10-30
Remove endpoints that create/update/delete archived Profile resources

### LINKED ISSUE #4605: Task - Applicant - remove/disable routes used by archived profile forms
# Description
We've archived several pages on the Applicant Profile. We've disabled the submit buttons of the forms on those pages, but the 'disabled' attribute on the buttons can simply be removed and the forms remain otherwise functional.

To avoid any accidental or malicious updating of data which we no longer expect to change, we should disable or remove any server routes which the archived forms use to create, update, or delete data.

Stripping out some old code too.

tristan-orourke:
Resolves #4605.

Removes the endpoints used by the forms on archived Applicant profile pages, along with any controller functions, validators, and tests reached exclusively by those endpoints.

Also, the policies for the models associated with the archived profile pages have been changed so that no users (excepting admins) have permission to create, update or delete those models.


resources/views/applicant/strategic_response_application/common/skill-form.html.twig
```diff
@@ -79,7 +79,7 @@
         {% if locale == "application" and application_step == "4" and criterion == false %}template{% endif %}
         {% if skill_declaration %}complete{% endif %}
         modal-target-object"
-    data-item-saved="{{ skill_declaration ? "true" : "" }}" data-item-url="{{ route("skill_declarations.destroy", ":id") }}" data-item-id="{{ skill_declaration ? skill_declaration.id : 0 }}">
+    data-item-saved="{{ skill_declaration ? "true" : "" }}" data-item-url="" data-item-id="{{ skill_declaration ? skill_declaration.id : 0 }}">
```
yonikid15, line 82: I believe we can remove this entire folder `resources/views/applicant/strategic_response_application`. Since it's all been moved to its own app.

yonikid15 (approved):
Looks good 👍. 
@tristan-orourke We should also make an issue for the new Applicant Profile view for Managers. It's currently set to show the questions answered from "Your Narrative" section of the profile, which is getting archived.

resources/views/applicant/strategic_response_application/common/skill-form.html.twig
tristan-orourke: Created an issue https://github.com/GCTC-NTGC/TalentCloud/issues/4701

yonikid15: Nice, thanks 👍

PR #4654, opened 2020-11-05
Task - Timeline - Final submit fully functional

### LINKED ISSUE #4432: Task - Application - Make Final Submission fully functional
The submit button on the Final Submission step should be disabled unless the Application is already valid for submission (minus the signature and date fields).

Clicking the submit button saves the signature and date to the server and triggers a submission attempt on the server. 

If backend validation passes, the application status should be changed to 'submitted', a snapshot must be saved, and when the request returns successfully the frontend should navigat to the Congrats page. 

If validation fails, or the request fails for any other reason, the frontend should display the error and remain on the page.

yonikid15:
Resolves #4432 

### Notes:
- The error messages displayed after a failed submission need to be tweaked to better inform the applicant which step the error is coming from. However, that should probably be moved to a new issue.
image
The way the timeline is setup though they shouldn't be able to click "Submit Application" unless all the previous steps have been completed.


routes/web.php
```diff
@@ -1070,6 +1070,10 @@ function (): void {
         ->where('application', '[0-9]+')
         ->middleware('can:view,application')
         ->name('application.basic.update');
+    Route::put('applications/{application}/submit', 'Api\ApplicationController@submit')
+        ->where('application', '[0-9]+')
+        ->middleware('can:view,application')
```
tristan-orourke, line 1075: I think this should require permission to update, not view.

tristan-orourke (changes requested):
Overall looks good, just a few specific comments.

app/Services/Validation/ApplicationTimelineValidator.php
```diff
@@ -330,6 +330,23 @@ public function validateComplete(JobApplication $application)
 
     /* Step 6 ------------------------------------------------------------------------------------------------------- */
 
+    /**
+     * Affirmation Rules.
+     * @var mixed
+     */
+    public $affirmationRules = [
```
tristan-orourke, line 337: Now that these rules have been extracted like this, make sure to reference this in the affirmationValidator function, instead of copy-pasting the rules.

tristan-orourke (changes requested):

tristan-orourke (approved):

PR #4664, opened 2020-11-06
Replace loading state icon with blinking animation

yonikid15:
### Notes:
- The loading state was only icon which prevented the user from clicking on it to try and reach another step. This 
removes the loading state and adds a blinking animation instead.


tristan-orourke (approved):
Love the new effect! It looks great, feels nice to use.

PR #4678, opened 2020-11-09
Update routes and links for new Applications

### LINKED ISSUE #4647: Task - Application - Finalize new Application urls
# Description
Ensure urls and routes are set up so that:
- When a user clicks 'Apply Now' a version2 application is created, and they are redirected to the Welcome page for that application
- When a user clicks to edit a draft version2 application, they are directed to the last touched step for that applicaiton
- When a user clicks to edit a draft version1 application, they are directed to the old application process
- When a user (Manager or Applicant) clicks to view a Complete application, they are taken to the correct view for version1 _and_ version2 applications
- When a user clicks a link on the Review step of a version2 application, they are taken to the correct step (via SPA-style navigation)

tristan-orourke:
Resolves #4647.

### Notes:
- URL to application timeline has been moved out of demo section. Now, go to `/applications/{applicationId}/edit`.
- Clicking 'apply' on a job poster now creates a version2 application and brings you to new UI.
- If you access version 1 application for a job that's still open, you will be sent to old application UI. I'm not sure we want to support this at all, since there really shouldn't be any extant jobs in production. Should we remove the old application UI (and routes) entirely?
- Wrote a test for a new controller function (JobsController@apply) which routes applicants to the right page and creates a new Application if necessary.
- Moved 'Next Steps' route of demo section to `/applications/{applicationId}/next`
- With these urls, the default breadcrumbs link back to Application index, not the Job Poster. In fact, there's no easy way to go back and review the job poster you're applying for. Is that something we want to address with breadcrumbs, or some other way?


resources/views/applicant/application_index/application.html.twig
```diff
@@ -57,7 +57,16 @@
             <div data-c-grid-item="base(1of2) tl(1of6)" data-c-align="base(center)">
                 {% if item.application_status.name == "draft" %}
                     {% set expired = item.job_poster.close_date_time.isPast() %}
-                    <a data-c-button="solid(c1)" data-c-radius="rounded" href={{ expired ? route('applications.show', item) : route('job.application.edit.1', item.job_poster) }} title="{{ expired ? application_index.view_link_title : application_index.draft_link_title }}">{{ expired ? application_index.view_link_label : application_index.draft_link_label }}</a>
+                    <a
+                        data-c-button="solid(c1)"
+                        data-c-radius="rounded"
+                        href={{ expired
+                            ? route('applications.show', item)
+                            : item.version_id is same as(2)
```
Khristophor, line 65: Probably fine as-is, but might be worth passing the `$version2id` through to the view for comparison here

resources/views/applicant/application_index/application.html.twig
```diff
@@ -16,7 +16,7 @@
             </div>
             <div data-c-grid-item="tl(1of3)" data-c-align="base(left) tl(right)">
                 {# Status #}
-                {% if item.application_status.name != "draft"  %}
+                {% if not item.isDraft()  %}
```
Khristophor, line 19: This kind of thing we need to be careful with (both the old and new syntax) as it's running the query on the application_status table for every loop iteration. We should add an eager load in the Controller or a `with` array in the model itself

Khristophor (approved):
@tristan-orourke Looks good! Might be worth splitting off an issue for the duplicate query thing in the template for @mnigh to pick up and gain familiarity with.

Khristophor:
> With these urls, the default breadcrumbs link back to Application index, not the Job Poster. In fact, there's no easy way to go back and review the job poster you're applying for. Is that something we want to address with breadcrumbs, or some other way?

Also probably worth getting some UX input on this... I know if I was going through our multi-step application process I would appreciate being able to either preview the poster or be able to jump back and forth.

resources/views/applicant/application_index/application.html.twig
tristan-orourke: Good call. I'll set up an issue for this.

PR #4706, opened 2020-11-16
Avoid extra data fetches in JPB and improve navigation between steps

### LINKED ISSUE #4613: Task - Job Poster Builder - Don't re-fetch data that's kept in Redux store every step
# Description
We keep data like the Job poster in the redux store persisted across all steps of the Job Poster Builder, but the JobBuilderStep component re-fetches it every step anyway. We should avoid re-fetching any data we already have. 

In addition, the links in the Progress Bar to different steps are disabled until this refetched data finishes loading. If we have a previous version of the necessary data available in the redux store, we should use that to construct the links even while the data is loading. Having the links available instantly will improve the perceived performance of JPB pages. (We could change this to match the newer Application progress bar, where the step icons blink while data is loading, instead of being un-clickable spinners.)

tristan-orourke:
Resolves #4613.

### Notes:
- Rewrites data fetching functions to hooks, modeled after the more recent applicationHooks. 
- Avoids re-fetching job, tasks and criteria when navigating between steps of Job Poster Builder.
- Allows the links in progress bar to be clicked even while data is updating, as long as they've previously been visited.

### DIFF
--- a/resources/assets/js/components/JobBuilder/JobBuilderProgressTracker.tsx
+++ b/resources/assets/js/components/JobBuilder/JobBuilderProgressTracker.tsx
@@ -58,8 +58,9 @@ const stepComesBefore = (
   return comesBefore;
 };
 
-export const JobBuilderProgressTracker: React.FunctionComponent<JobBuilderProgressTrackerProps &
-  WrappedComponentProps> = ({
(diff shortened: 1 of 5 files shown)


Khristophor (approved):
Nice, those hooks will be super handy

PR #4708, opened 2020-11-17
Task - Experience skills bugfix + add soft deletes task

### LINKED ISSUE #4615: Application - Experience Step - Saving experience deletes ExperienceSkills not relevant to current job
# Description
When Experience is saved, requests are sent to delete any ExperienceSkills that are no longer checked. However, this currently includes skills for which no checkbox even appears, because they aren't relevant to the current job.

yonikid15:
Resolves #4615

### DIFF
--- a/app/Http/Controllers/Api/ExperienceSkillsController.php
+++ b/app/Http/Controllers/Api/ExperienceSkillsController.php
@@ -16,7 +16,8 @@ class ExperienceSkillsController extends Controller
         $softDeletedExperienceSkill = ExperienceSkill::onlyTrashed()
             ->where([
             ['skill_id', $data['skill_id']],
-            ['experience_id', $data['experience_id']]
+            ['experience_id', $data['experience_id']],
+            ['experience_type', $data['experience_type']]
         ])->first();
         if ($softDeletedExperienceSkill) {
             $softDeletedExperienceSkill->restore();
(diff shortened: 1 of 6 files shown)


app/Http/Controllers/Api/ExperienceSkillsController.php
```diff
@@ -12,13 +12,26 @@ class ExperienceSkillsController extends Controller
...
-        $experienceSkill->justification = $data['justification'];
-        $experienceSkill->save();
-        return new JsonResource($experienceSkill->fresh());
+        // Restore soft deleted experienceSkill if it exists, otherwise create a new one.
+        $softDeletedExperienceSkill = ExperienceSkill::onlyTrashed()
+            ->where([
+            ['skill_id', $data['skill_id']],
+            ['experience_id', $data['experience_id']],
+            ['experience_type', $data['experience_type']]
+        ])->first();
```
Khristophor, line 21: Assuming the ExperienceSkill ID is returned with the `$request` it's probably safer/faster to use that to check for soft deleted models

Khristophor (changes requested):

app/Http/Controllers/Api/ExperienceSkillsController.php
yonikid15: I agree, however it's not. The id is auto incremented when a new experience skill is created.

Khristophor: Ah gotcha. Sorry I thought `store` was being used for creates and updates.

Khristophor (approved):

PR #4772, opened 2020-11-23
Create Skill Category table and model

### LINKED ISSUE #4690: Task - Create Skill Category table and model
Create the SkillCategory table & model.
Each SkillCategory needs a key, a name in both languages, and optionally specifies a parent SkillCategory (which means they could be traversed as a tree).

Something about ordering.

mnigh:
Resolves #4690


database/migrations/2020_11_23_165933_create_skill_categories_table.php
```diff
@@ -0,0 +1,34 @@
...
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::create('skill_categories', function (Blueprint $table) {
+            $table->increments('id');
+            $table->string('key');
+            $table->json('name');
+            $table->integer('parent_category_id')->default(0)->nullable();
```
tristan-orourke, line 20: Since its nullable, leave the default as null.

app/Models/SkillCategory.php
```diff
@@ -0,0 +1,39 @@
...
+use Backpack\CRUD\app\Models\Traits\CrudTrait;
+use Backpack\CRUD\app\Models\Traits\SpatieTranslatable\HasTranslations;
+
+/**
+ * Class SkillCategory
+ *
+ * @property int $id
+ * @property string $key
+ * @property string $name
+ * @property int $parent_category_id
```
tristan-orourke, line 14: The type should be int|null

tristan-orourke (changes requested):

database/migrations/2020_11_23_165933_create_skill_categories_table.php
mnigh: @tristan-orourke Good catch, thanks. Looking at this again, I wonder if it's better to remove nullable and leave the default as 0 so that the model doesn't have a mixed type: thoughts?

tristan-orourke: I actually prefer using null, its more unambiguous than having to know 0 represents no parent.

mnigh: @tristan-orourke The [documentation](https://backpackforlaravel.com/docs/4.0/crud-operation-reorder#requirements) for Backpack's Reorder feature (which is implemented in #4811) says to use a default value of 0. Did you want me to use a default value of null anyway?

tristan-orourke (changes requested):

database/migrations/2020_11_23_165933_create_skill_categories_table.php
tristan-orourke: Okay, if we're using the reorder feature we should use their defaults.
I find it odd that they want to default to 0 but also make it nullable, so just leave a comment here and in the model referring to the Backpack feature.

tristan-orourke (approved):

tristan-orourke (approved):

PR #4785, opened 2020-11-24
Task - batch request experience skills on experience step

### LINKED ISSUE #4660: Task - Application - Speed up Saving on Experience step by updating ExperienceSkills in batches
# Description
When saving an Experience, every skill which is added or removed causes a new request to Create/Update/Delete an individual ExperienceSkill entity. Updating the Experience isn't finished until all these requests return. It would feel much faster if all the skills could be updated in a single batch request.

yonikid15:
Resolves #4660 

### Notes:
- I feel like the current authorization could be improved, let me know if you have any suggestions.


tests/Feature/Api/ExperienceSkillsControllerTest.php
```diff
@@ -124,4 +124,88 @@ public function testRestoreSoftDeletedExperienceSkill()
...
+        $faker = \Faker\Factory::create();
+        $work = factory(ExperienceWork::class)->create();
+        $experienceSkill1 = factory(ExperienceSkill::class)->create([
+            'experience_id' => $work->id,
+            'experience_type' => 'experience_work',
+        ]);
+        $experienceSkill2 = factory(ExperienceSkill::class)->create([
+            'experience_id' => $work->id,
+            'experience_type' => 'experience_work',
+        ]);
```
tristan-orourke, line 162: Maybe explicitly set skill ids here? You could conceivably get the same skill twice (though unlikely).

tests/Feature/Api/ExperienceSkillsControllerTest.php
```diff
@@ -124,4 +124,88 @@ public function testRestoreSoftDeletedExperienceSkill()
...
+            collect($experienceSkill1->attributesToArray())->forget(['justification', 'updated_at'])->all(),
+            ['justification' => $updateData[0]['justification']],
+            collect($experienceSkill2->attributesToArray())->forget(['justification', 'updated_at'])->all(),
+            ['justification' => $updateData[1]['justification']]
+        );
+        $response->assertJsonFragment($expectData);
+        $this->assertDatabaseHas('experience_skills', $expectData);
+    }
+
+    public function testBatchDeleteExperienceSkill()
```
tristan-orourke, line 183: One other test you could do would be to confirm that updating/creating a skill after soft deleting just un-soft-deletes it, instead of creating a brand new skill.

tristan-orourke (changes requested):

app/Http/Controllers/Api/ExperienceSkillsController.php
```diff
@@ -48,4 +49,67 @@ public function destroy(ExperienceSkill $experienceSkill)
...
+        foreach ($newExperienceSkills as $newExperienceSkill) {
+            // Restore soft deleted experienceSkill if it exists, otherwise create a new one.
+            $softDeletedExperienceSkill = ExperienceSkill::onlyTrashed()
+                ->where([
+                ['skill_id', $newExperienceSkill['skill_id']],
+                ['experience_id', $newExperienceSkill['experience_id']],
+                ['experience_type', $newExperienceSkill['experience_type']]
+            ])->first();
+            if ($softDeletedExperienceSkill) {
+                $softDeletedExperienceSkill->restore();
```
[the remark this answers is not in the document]
tristan-orourke: Check that its not an empty string, as well.

app/Http/Requests/BatchStoreExperienceSkill.php
```diff
@@ -0,0 +1,83 @@
...
+                    $experience = ExperienceEducation::find($experience_id);
+                    break;
+                case 'experience_personal':
+                    $experience = ExperiencePersonal::find($experience_id);
+                    break;
+            }
+
+            if ($experience === null ||
+                $user === null ||
+                !$user->can('create', ExperienceSkill::class) ||
```
tristan-orourke, line 50: The 'create' check only needs to happen once, not in the loop.

app/Http/Requests/BatchStoreExperienceSkill.php
```diff
@@ -0,0 +1,83 @@
...
+                !$user->can('update', $experience)
+            ) {
+                return false;
+            }
+        }
+
+        return $result;
```
tristan-orourke, line 57: Don't need $result, can just return true;

routes/web.php
```diff
@@ -1003,6 +1003,15 @@ function (): void {
         ->middleware('can:delete,experienceSkill')
         ->name('experience-skill.destroy');
 
+    Route::post('experience-skills/batch-store', 'Api\ExperienceSkillsController@batchStore')
+        ->middleware('can:create,App\Models\ExperienceSkill')
+        ->name('experience-skill.batch-store');
+    Route::put('experience-skills/batch-update', 'Api\ExperienceSkillsController@batchUpdate')
+        ->name('experience-skill.batch-update');
+    Route::put('experience-skills/batch-destroy', 'Api\ExperienceSkillsController@batchDestroy')
+        ->name('experience-skill.batch-destroy');
```
tristan-orourke, line 1012: I think we should make these both POST requests, since they don't really match a REST api design, and don't really reflect the normal PUT use case.

app/Http/Controllers/Api/ExperienceSkillsController.php
```diff
@@ -48,4 +49,70 @@ public function destroy(ExperienceSkill $experienceSkill)
     {
         $experienceSkill->delete();
     }
+
+    public function batchStore(BatchStoreExperienceSkill $request)
+    {
+        $newExperienceSkills = $request->validated();
+        $response = [];
+
+        foreach ($newExperienceSkills as $newExperienceSkill) {
```
tristan-orourke, line 58: @Khristophor had some good suggestions on how to deal with batch requests. One thing would be to wrap this for loop (and similarly in the other batch requests) in a [DB::transaction](https://laravel.com/docs/8.x/database#database-transactions) to ensure we don't end up with a partially-failed-partially-successful request.

app/Http/Controllers/Api/ExperienceSkillsController.php
```diff
@@ -48,4 +49,70 @@ public function destroy(ExperienceSkill $experienceSkill)
...
+    {
+        $newExperienceSkills = $request->validate([
+            '*.id' => 'required|exists:App\Models\ExperienceSkill,id',
+            '*.justification' => 'nullable|string',
+        ]);
+        $response = [];
+
+        foreach ($newExperienceSkills as $newExperienceSkill) {
+            $experienceSkill = ExperienceSkill::where('id', $newExperienceSkill['id'])->first();
+            $this->authorize('update', $experienceSkill);
```
tristan-orourke, line 98: Let's move this authorization to a Form Request, like with the batch store, for consistency.

tristan-orourke (changes requested):

app/Http/Controllers/Api/ExperienceSkillsController.php
```diff
@@ -48,4 +49,70 @@ public function destroy(ExperienceSkill $experienceSkill)
...
+    public function batchUpdate(Request $request)
+    {
+        $newExperienceSkills = $request->validate([
+            '*.id' => 'required|exists:App\Models\ExperienceSkill,id',
+            '*.justification' => 'nullable|string',
+        ]);
+        $response = [];
+
+        foreach ($newExperienceSkills as $newExperienceSkill) {
+            $experienceSkill = ExperienceSkill::where('id', $newExperienceSkill['id'])->first();
```
tristan-orourke, line 97: @Khristophor also had some optimization suggestions for this. You can get all the ExperienceSkills with one query outside of the loop with `ExperienceSkills::whereIn('id', $newExperienceSkills ->pluck('id')->all())->get()`. We want to minimize direct DB manipulation within the loop.

app/Http/Controllers/Api/ExperienceSkillsController.php
```diff
@@ -48,4 +49,70 @@ public function destroy(ExperienceSkill $experienceSkill)
...
+                $softDeletedExperienceSkill->save();
+                array_push($response, $softDeletedExperienceSkill);
+            } else {
+                $experienceSkill = new ExperienceSkill($newExperienceSkill);
+                $experienceSkill->skill_id = $newExperienceSkill['skill_id'];
+                $experienceSkill->experience_id = $newExperienceSkill['experience_id'];
+                $experienceSkill->experience_type = $newExperienceSkill['experience_type'];
+                $experienceSkill->justification = $newExperienceSkill['justification'];
+                $experienceSkill->save();
+                $experienceSkill->fresh();
```
tristan-orourke, line 80: May be able to avoid calling ->fresh() inside the loop, and refresh all the objects with a single DB query by with `return new JsonResource($response->fresh());` at the end.

Khristophor: This will only work if it's an Eloquent collection, though. Not sure if it's doable here where we're pushing to an empty array.

app/Http/Controllers/Api/ExperienceSkillsController.php
```diff
@@ -48,4 +49,70 @@ public function destroy(ExperienceSkill $experienceSkill)
...
+                array_push($response, $experienceSkill);
+            }
+        }
+
+        return new JsonResource($response);
```
Khristophor, line 85: I also tend to use `return JsonResource::collection($response);` when it's an array of items rather than a single object being returned. Not sure if it actually makes a difference but it's [outlined in the docs here](https://laravel.com/docs/8.x/eloquent-resources#resource-collections).

app/Http/Controllers/Api/ExperienceSkillsController.php
```diff
@@ -48,4 +49,70 @@ public function destroy(ExperienceSkill $experienceSkill)
...
+            '*.justification' => 'nullable|string',
+        ]);
+        $response = [];
+
+        foreach ($newExperienceSkills as $newExperienceSkill) {
+            $experienceSkill = ExperienceSkill::where('id', $newExperienceSkill['id'])->first();
+            $this->authorize('update', $experienceSkill);
+            $experienceSkill->fill($newExperienceSkill);
+            $experienceSkill->save();
+            array_push($response, $experienceSkill);
```
Khristophor, line 101: For the batch update I think you can forego the array and just `return JsonResource::collection($experienceSkills->fresh());` (assuming you grab them all with a single query like @tristan-orourke mentioned above).

yonikid15:
@tristan-orourke @Khristophor I think I covered everything mentioned above, let me know If I missed something.

Khristophor (approved):
Looking good to me, @tristan-orourke want to confirm?

tristan-orourke (approved):

PR #4802, opened 2020-11-25
Task - batch request experience skills on skills step

### LINKED ISSUE #4662: Task - Application - Speed up Save All on Skills step by submitting all ExperienceSkills in a single batch request
# Description
Right now each ExperienceSkill form starts it own request to the api. We could speed this up and limit the number of requests if we allowed multiple ExperienceSkills to be updated in a single batch request.

yonikid15:
Resolves #4662 

### Notes:
- Branches off of #4785


resources/assets/js/store/Experience/experienceActions.ts
```diff
@@ -308,6 +312,108 @@ export const deleteExperienceSkill = (
...
+export const batchUpdateExperienceSkills = (
+  experienceSkills: ExperienceSkill[],
+): RSAActionTemplate<
+  typeof BATCH_UPDATE_EXPERIENCE_SKILLS_STARTED,
+  typeof BATCH_UPDATE_EXPERIENCE_SKILLS_SUCCEEDED,
+  typeof BATCH_UPDATE_EXPERIENCE_SKILLS_FAILED,
+  ExperienceSkill[],
+  {}
+> =>
+  asyncPut(
```
tristan-orourke, line 373: I think we should make all the Batch endpoints POST, as mentioned here https://github.com/GCTC-NTGC/TalentCloud/pull/4785/files#r536387663

resources/assets/js/store/Experience/experienceActions.ts
```diff
@@ -308,6 +312,108 @@ export const deleteExperienceSkill = (
...
+export const batchDeleteExperienceSkills = (
+  experienceSkills: ExperienceSkill[],
+): RSAActionTemplate<
+  typeof BATCH_DELETE_EXPERIENCE_SKILLS_STARTED,
+  typeof BATCH_DELETE_EXPERIENCE_SKILLS_SUCCEEDED,
+  typeof BATCH_DELETE_EXPERIENCE_SKILLS_FAILED,
+  {},
+  ExperienceSkill[]
+> =>
+  asyncPut(
```
tristan-orourke, line 407: I think we should make all the Batch endpoints POST, as mentioned here https://github.com/GCTC-NTGC/TalentCloud/pull/4785/files#r536387663

resources/assets/js/store/Experience/experienceReducer.ts
```diff
@@ -493,6 +513,16 @@ export const entitiesReducer = (
           action.meta.experienceType,
         ),
       };
+    case BATCH_CREATE_EXPERIENCE_SKILLS_SUCCEEDED:
+    case BATCH_UPDATE_EXPERIENCE_SKILLS_SUCCEEDED:
```
tristan-orourke, line 517: Careful here. The Batch Update and Create requests currently return the list of _updated_ ExperienceSkills, not _all_ ExperienceSkills. I believe `setExperienceSkills` rewrites the entire list with just the payload of this action. We don't actually don't want to delete any experienceSkills which weren't part of these batch requests.

tristan-orourke (changes requested):

resources/assets/js/store/Experience/experienceReducer.ts
yonikid15: Actually, I believe `setExperienceSkills` is working as intended since it takes the previous `state` and spreads the `experienceSkills.byId` list with the entire list (including the ones that weren't part of the payload), then adds/overwrites the created/updated values that were part of the request.

app/Http/Requests/BatchStoreExperienceSkill.php
```diff
@@ -0,0 +1,85 @@
...
+                    break;
+            }
+
+            if ($experience === null ||
+                $user === null ||
+                !$user->can('update', $experience)
+            ) {
+                return false;
+            }
+        }
```
Khristophor, line 57: I think this section could be reworked as well. The ExperienceSkill -> Experience relationship is polymorphic, so we should be able to just grab all the ExperienceSkills from the database. Something like
```php
$inputIds = collect($this->all())->pluck('id')->all();
$experienceSkills = ExperienceSkill::with('experience')->whereIn('id', $inputIds)->get();

foreach ($experienceSkills as $experienceSkill) {
  if ($experience === null ||
    $user === null ||
    !$user->can('update', $experienceSkill->experience())
  ) {
    return false;
  }
}
```

yonikid15: Thanks! This solution works for the batch updating so I added it into the new Form Request I made. However, for batch storing this can't work, `$experienceSkills` will always be empty since it doesn't exist in the DB yet.

Khristophor: Gotcha, I noticed that on the other batch-update PR

resources/assets/js/store/Experience/experienceReducer.ts
tristan-orourke: Somehow this ended up in the other PR, but you're right. `setExperienceSkills` looks like it adds to the state, not replaces it like I originally thought.

tristan-orourke (approved):

PR #4799, opened 2020-11-25
Add Skill Category API

mnigh:
Resolves #4692.

### Notes:
- Requires #4772 and #4789  to be merged first.

PR #4789, opened 2020-11-24
Create Skills, Skills Category relationship table

mnigh:
Resolves #4691

### Notes:
- Requires #4772 to be merged first.

PR #4811, opened 2020-11-26
Create admin interfaces for Skill Categories and linking them to Skills

### LINKED ISSUE #4716: Task - Create Admin interfaces for creating new Skill Categories and linking them to Skills
Create a page on Admin portal for creating and editing SkillCategories. Note that when editing a SkillCategory, the parent field should be represented by a dropdown menu of all other existing SkillCategories. We should have some validation on the backend that prevents a category having itself as its parent, or from creating a cyclic relationship (eg A has parent B, B has parent C, C has parent A).

Also add a multi-select field to Skills on admin portal, which connects them to multiple SkillCategories.

Needs some validation: Parents can't have skills related to them, can't relate a category to itself.

Also need to handle ordering.

mnigh:
Resolves #4716.

### Notes:
- Requires #4772, #4789, and #4799 to be merged first.


app/Http/Controllers/Admin/SkillCrudController.php
```diff
@@ -66,6 +67,20 @@ public function setup() : void
...
+            $this->crud->addField([
+                'name' => 'skill_categories',
+                'type' => 'select2_multiple',
+                'label' => 'Skill Categories (select all that apply)',
+                'entity' => 'skills',
+                'attribute' => 'name',
+                'model' => 'App\Models\SkillCategory',
+                'pivot' => true,
+                'options' => (function ($query) {
+                    // Exclude skill categories with no children.
```
tristan-orourke, line 79: This should be `with no parent`, yes? 😛

tristan-orourke (approved):

PR #4812, opened 2020-11-26
Bugfix - application validator test

### LINKED ISSUE #4639: Bug - Application Skill validator test failing intermittently
# Description
Two unit tests fail sporadically causing build issues on PRs and dependency updates:

```
1) Tests\Unit\Validators\StrategicResponseApplicationValidatorTest::testSkillsComplete

Failed asserting that true is false.

/var/www/tests/Unit/Validators/StrategicResponseApplicationValidatorTest.php:174

-------------------------------------------------------------------------------------

1) Tests\Unit\Validators\ApplicationValidatorTest::testSkillsComplete

Failed asserting that true is false.

/var/www/tests/Unit/Validators/ApplicationValidatorTest.php:120
```

either both tests are bad, the skill validator is bad, or we have some factory generation that falls outside the expected data structure and throws a failure randomly.

Reproducing locally has been problematic (I ran these tests for over an hour one day without a failure).

yonikid15:
Resolves #4639 

### Notes:
- Tried switching to using the words method in faker. However, we'll need to get Travis back up (or switch to Git Actions) before finding out if it worked.


database/factories/SkillDeclarationFactory.php
```diff
@@ -15,6 +15,6 @@
             return factory(Applicant::class)->create()->id;
         },
         'skillable_type' => 'applicant',
-        'description' => $faker->text(100),
+        'description' => $faker->words(50, true),
```
Khristophor, line 18: This will actually be prone to occasional failure. `text()` has a hard limit, while `words()` will generate 50 words of a varying number of characters.

It was set to `$faker->paragraphs(3, true),` before, but I swapped it to `text()`

[More deets here](https://github.com/fzaninotto/Faker#fakerproviderlorem)

Khristophor (changes requested):
I don't think the max character count for the description field is the issue. I tracked down a couple failing Travis builds with files/line numbers to look into:

https://travis-ci.com/github/GCTC-NTGC/TalentCloud/jobs/403784551
https://travis-ci.com/github/GCTC-NTGC/TalentCloud/jobs/400258600
https://travis-ci.com/github/GCTC-NTGC/TalentCloud/jobs/398817821
https://travis-ci.com/github/GCTC-NTGC/TalentCloud/jobs/398018736

Most of those are Strategic Response Application Validator tests, so you can ignore them if you think the logic isn't the same in the old Application Validator test. The first one is for the new TimelineValidator, but it might not be relevant anymore.

tl;dr most failures seem to come from the StrategicResponseValidator, if you're just deleting that, we might be fine

database/factories/SkillDeclarationFactory.php
yonikid15: That's what we want though right, since we're using the WordLimitRule at max 100 words

Khristophor: Ah okay, I thought it was a 100 character limit. Either way though, 100 characters won't fail a 100 word count validation check, so I don't think how it was would be prone to random failures.

yonikid15: Ya, that's true. Well hopefully removing the Strategic Response Validator will fix it.

tristan-orourke:
Well, this passed the tests the first time at least. Probably good to merge.

tristan-orourke (approved):

Khristophor (approved):

PR #4813, opened 2020-11-26
Create a Codecov Github Action

tristan-orourke:
Adds Codecov's "Github Action" to our automated workflows, to replace what we ran on Travis-CI.

We can test it on this pr. Once merged, it should run on all PR's into dev.


tristan-orourke:
I've just realized that Codecov doesn't actually create the coverage reports, and needs to be run after the unit tests 😂

PR #4814, opened 2020-11-26
Create a Github Actions workflow for CI tests

### LINKED ISSUE #4840: Task - Replace Travis CI tests with Github Actions
# Description
We've run out of free Travis credits and we need an alternate way to run CI tests for pull requests. [Github Actions](https://docs.github.com/en/free-pro-team@latest/actions) can fully replace the same functionality for free.

tristan-orourke:
Resolves #4840.

This should replace our use of Travis-CI with github actions.

We can test it on this branch. Once merged in, it should run on all PRs.


Khristophor:
I'm not sure you can set up the environment, and then run the tests as two different actions. I think each job is a distinct environment and can't be reused by other jobs.

.github/workflows/tests.yml
```diff
@@ -36,11 +36,9 @@ jobs:
           ${{ runner.os }}-php-
     - name: Install dependencies
       if: steps.composer-cache.outputs.cache-hit != 'true'
-      uses: php-actions/composer@v4
+      uses: MilesChou/composer-action/5.5/install@master
```
mnigh, line 39: @tristan-orourke Shouldn't this be using PHP version 7.2 instead pf 5.5?

`uses: MilesChou/composer-action/7.2/install@master`

tristan-orourke: Oops! Copy paste errors! Thanks.

.github/workflows/tests.yml
```diff
@@ -96,10 +96,14 @@ jobs:
       with:
         name: laravel-logs
         path: storage\logs\*.log
-    - name: Slack Notification
-      uses: rtCamp/action-slack-notify@v2
-      env:
-        SLACK_WEBHOOK: ${{ secrets.SLACK_WEBHOOK }}
+     - name: Send Slack notification
+       uses: 8398a7/action-slack@v2
```
mnigh: @gggrant Any reason to use v2 instead of [v3 (latest)](https://github.com/8398a7/action-slack) for the Slack notifications?

.github/workflows/tests.yml
```diff
@@ -132,7 +136,11 @@ jobs:
        with:
          files: report/jest/clover.xml,report/phpunit/junit.jest.xml
          fail_ci_if_error: true
-     - name: Slack Notification
-       uses: rtCamp/action-slack-notify@v2
+     - name: Send Slack notification
+       #if: failure()
+       uses: 8398a7/action-slack@v2
```
mnigh, line 170: @gggrant Any reason to use v2 instead of [v3 (latest)](https://github.com/8398a7/action-slack) for the Slack notifications?

.github/workflows/tests.yml
gggrant: Nope! I'll update when doing the browser tests.

.github/workflows/tests.yml
```diff
@@ -0,0 +1,200 @@
...
+    steps:
+    - uses: actions/checkout@v2
+    - uses: shivammathur/setup-php@v2
+      with:
+        php-version: 7.2
+        extensions: dom, curl, libxml, mbstring, zip, pcntl, pdo, bcmath, soap, intl, gd, exif, iconv, imagick
+        tools: composer:v1
+
+    - name: Copy .env
+      run: cp .env.travis .env
```
Khristophor, line 34: Should add a new `.env` for Github actions, or rename it, for the sake of not confusing ourselves and others in 8 months

.github/workflows/tests.yml
```diff
@@ -0,0 +1,200 @@
...
+    - name: Install dependencies
+      if: steps.composer-cache.outputs.cache-hit != 'true'
+      uses: MilesChou/composer-action/7.2/install@master
+      with:
+        args: install --no-interaction --prefer-dist --no-progress --no-suggest
+
+    - name: Setup Node.js
+      uses: actions/setup-node@v1
+      with:
+        node-version: 12.6.0
```
Khristophor, line 54: I think [14 is the current LTS version](https://nodejs.org/en/). Can we pin this to LTS so things don't break as we make future updates?

Khristophor (changes requested):

.github/workflows/tests.yml
```diff
@@ -0,0 +1,277 @@
...
+    - name: Generate certificates
+      run: make gen-certs
+
+    - name: Create Laravel storage symlink
+      run: php artisan storage:link
+
+    - name: Setup database
+      run: |
+        docker exec -d postgres psql -c 'create database testing with owner talentcloud;' -U talentcloud --no-password
+        php artisan migrate --seed
```
tristan-orourke, line 87: I think this can be removed to save a bit of time. Most of the phpunit tests reset the DB anyway.

.github/workflows/tests.yml
```diff
@@ -0,0 +1,277 @@
...
+      env:
+        DB_PORT: 5432
+        DB_HOST: localhost
+        DB_DATABASE: testing
+
+    - name: Compile webpack assets
+      run: npm run prod
+
+    - name: Start a web server
+      run: php artisan serve &
```
tristan-orourke, line 97: I'm not sure if this is necessary in the end. We should see if the tests can pass without it.

.github/workflows/tests.yml
```diff
@@ -0,0 +1,277 @@
...
+      with:
+        files: report/phpunit/coverage.xml,report/phpunit/junit.phpunit.xml
+        fail_ci_if_error: true
+
+    - name: Archive Laravel logs
+      if: ${{ always() }}
+      uses: actions/upload-artifact@v2
+      with:
+        name: laravel-logs
+        path: storage\logs\*.log
```
tristan-orourke, line 118: This step hasn't been finding any log files to store. If we can't fix that we might as well delete the step.

.github/workflows/tests.yml
Khristophor: They don't actually seed without explicit options in the unit test (which we don't have)

https://laravel.com/docs/8.x/database-testing#using-seeds

gggrant (approved):

PR #4871, opened 2020-12-03
Create relationship between Applicant and Skill

### LINKED ISSUE #4846: Task - Update data architecture to allow Skills to be attached directly to Applicants
# Description
Currently, Applicants are linked to Experiences and Experiences are linked to Skills. In addition to this, we want Applicants to link directly to Skills. 

For now, no other data needs to be part of this relationship, so it can be represented by a simple pivot table. There should be a uniqueness constraint ensuring an Applicant cannot have the same skill multiple times.

# Acceptance criteria
- [ ] Database is updated
- [ ] Model classes are updated

mnigh:
Resolves #4846.

### Notes:
- Creates a many-to-many relationship between the Applicant model and the Skill model.
- Adds a uniqueness constraint ensuring an applicant cannot have the same skill multiple times.

### DIFF
--- a/app/Models/Applicant.php
+++ b/app/Models/Applicant.php
@@ -124,7 +124,7 @@ class Applicant extends BaseModel
 
     public function skills() // phpcs:ignore
     {
-        return $this->belongsToMany(\App\Models\Skill::class, 'applicants_skills');
+        return $this->belongsToMany(\App\Models\Skill::class, 'applicant_skill');
     }
 
     // Version 2 application models.
--- a/app/Models/Applicant.php
+++ b/app/Models/Applicant.php
@@ -122,6 +122,11 @@ class Applicant extends BaseModel
         return $this->morphMany(\App\Models\Project::class, 'projectable');
     }
 
+    public function skills() // phpcs:ignore
+    {
+        return $this->belongsToMany(\App\Models\Skill::class, 'applicants_skills');
+    }
+
     // Version 2 application models.
 
     public function experiences_work() //phpcs:ignore
--- a/app/Models/Applicant.php
+++ b/app/Models/Applicant.php
@@ -29,6 +29,7 @@ use App\Traits\TalentCloudCrudTrait as CrudTrait;
  * @property \Illuminate\Database\Eloquent\Collection $references
  * @property \Illuminate\Database\Eloquent\Collection $work_samples
  * @property \Illuminate\Database\Eloquent\Collection $projects
+ * @property \Illuminate\Database\Eloquent\Collection $skills
  *
  * Version 2 application models.
  * @property \Illuminate\Database\Eloquent\Collection $experiences_work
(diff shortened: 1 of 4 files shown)


database/migrations/2020_12_03_151728_create_applicants_skills_table.php
```diff
@@ -0,0 +1,43 @@
...
+class CreateApplicantsSkillsTable extends Migration
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::create('applicants_skills', function (Blueprint $table) {
```
tristan-orourke, line 16: The default pivot table name should use the singular model names (applicant_skill). 
https://laravel.com/docs/8.x/eloquent-relationships#many-to-many-table-structure

database/migrations/2020_12_03_151728_create_applicants_skills_table.php
```diff
@@ -0,0 +1,43 @@
...
+            $table->increments('id');
+            $table->integer('applicant_id');
+            $table->foreign('applicant_id')
+                ->references('id')
+                ->on('applicants')
+                ->onUpdate('CASCADE')
+                ->onDelete('NO ACTION');
```
tristan-orourke, line 23: I think this could be CASCADE. If the Applicant itself is deleted, applicant skills should be part of that.

app/Models/Applicant.php
```diff
@@ -122,6 +122,11 @@ public function projects()
         return $this->morphMany(\App\Models\Project::class, 'projectable');
     }
 
+    public function skills() // phpcs:ignore
```
tristan-orourke, line 125: Add this relationship to the class's docblock.

app/Models/Skill.php
```diff
@@ -95,6 +95,11 @@ public function experience_skills() // phpcs:ignore
         return $this->hasMany(\App\Models\ExperienceSkill::class);
     }
 
+    public function applicants() // phpcs:ignore
```
tristan-orourke, line 98: Add this relationship to the class's docblock.

database/seeds/DevSeeder.php
```diff
@@ -218,5 +223,17 @@ public function run(): void
...
+            DB::table('applicants_skills')->updateOrInsert(
+                [
+                    'applicant_id' => Applicant::inRandomOrder()->get()->first()->id
+                ],
+                [
+                    'skill_id' => Skill::inRandomOrder()->get()->first()->id
+                ]
+            );
+        }
     }
```
tristan-orourke, line 238: I think it might be faster and more useful to focus on specific applicants here. Maybe just add 5 skills to $applicantUser->applicant. Also, the loop as is could accidentally add the same skill to the same applicant multiple times, causing a DB error.

tristan-orourke (changes requested):
A few changes to make, regarding conventions.

tristan-orourke (approved):

PR #4886, opened 2020-12-04
Attach Skill to Applicant during ExperienceSkill store

### LINKED ISSUE #4848: Task - API - When an ExperienceSkill is saved also save the Skill directly to Applicant
# Description
When an ExperienceSkill is saved, if the Skill is not yet attached to the experience's Applicant, then create that relationship.

This could be accomplished by doing some extra work in ExperienceSkillsController, or by using an [Event/Listener](https://laravel.com/docs/8.x/events) pattern and [listening for the saved event](https://laravel-news.com/laravel-model-events-getting-started) of the ExperienceSkill model.

Note: when an ExperienceSkill is deleted, there's no need to modify the Applicant-Skill relationship.

mnigh:
Resolves #4848 

### Notes:
- Requires #4871 to be merged first.


app/Http/Controllers/Api/ExperienceSkillsController.php
```diff
@@ -19,6 +19,17 @@ public function store(StoreExperienceSkill $request)
             ['experience_id', $data['experience_id']],
             ['experience_type', $data['experience_type']]
         ])->first();
+
+        // Attach skill to applicant if not already attached.
+        $applicant = $request->user()->applicant;
```
tristan-orourke, line 24: This is a bit of an edge case, but technically this might be the wrong applicant. An admin could use the api to create an ExperienceSkill attached to a different user's Experience.
We'll need to get the Experience object we're attaching to (see similar code in `Requests/StoreExperienceSkill.php`), then find the experience's Applicant. Note that depending on whether `$experience->experienceable_type` is 'applicant' or 'application', this is a little different: `$experience->experienceable` or `$experience->experienceable->applicant`.

There's already code in `Requests/StoreExperienceSkill.php` for getting an Experience from id and type. We want to use the same code here (and probably elsewhere in the future) so we should extract it a method somewhere. My first thought on where to put it is to create a common base class that all the Experience_____ models extend, and store some common utility methods there. (Actually, you could add another method there to extract getting Applicant from Experience, as well.)

app/Http/Controllers/Api/ExperienceSkillsController.php
```diff
@@ -19,6 +19,17 @@ public function store(StoreExperienceSkill $request)
...
+        $skillApplicantRelationshipExists = $applicant->skills()
+            ->wherePivot('applicant_id', $applicant->id)
+            ->wherePivot('skill_id', $data['skill_id'])
+            ->exists();
```
tristan-orourke, line 28: This seems overly verbose. Would this work equally well?
`$skillApplicantRelationshipExists = $applicant->skills->where('id', $data['skill_id'])->exists();`

app/Http/Controllers/Api/ExperienceSkillsController.php
```diff
@@ -19,6 +19,17 @@ public function store(StoreExperienceSkill $request)
...
+        if ($applicant != null && $skillApplicantRelationshipExists != 1) {
```
tristan-orourke, line 29: $skillApplicantRelationshipExists should be a boolean, no?

tests/Feature/Api/ExperienceSkillsControllerTest.php
```diff
@@ -124,4 +124,19 @@ public function testRestoreSoftDeletedExperienceSkill()
...
+    {
+        $work = factory(ExperienceWork::class)->create();
+        $workSkillData = $this->makeExpSkillData($work->id, 'experience_work');
+        $response = $this->actingAs($work->experienceable->user)
+            ->json('post', route('api.v1.experience-skill.store'), $workSkillData);
+        $response->assertOk();
+        $response->assertJsonFragment($workSkillData);
+        $id = $response->decodeResponseJson('id');
+        $this->assertDatabaseHas('applicant_skill', [
+            'applicant_id' => $work->experienceable->user->applicant->id,
```
tristan-orourke, line 138: You can skip a couple steps and just use `$work->experienceable_id`;

tristan-orourke (changes requested):

app/Http/Controllers/Api/ExperienceSkillsController.php
Khristophor: As a general practice, we also try to use strict equality checks (`===`, `!==`) whenever possible

tristan-orourke (approved):

PR #4923, opened 2020-12-11
Task - Applicant profile api

### LINKED ISSUE #4849: Task - API - Create Applicant Profile endpoints
# Description
Create API endpoints for getting and updating Applicant profile. This should include Citizenship and Veteran fields, and the list of linked Classifications.

# Images
This endpoint is currently designed to be used with the following page:
https://xd.adobe.com/view/f66135de-6ab1-449d-ac7c-98ade3ce1fd2-a332/screen/22903d0a-6389-4e0b-a349-94d757852618

yonikid15:
Resolves #4849 

### Notes:
- Requires #4869 to be merged first.


tristan-orourke (changes requested):
Sorry its been so long since this was waiting in review! It looks good, just needs some minor changes throughout to account for the fact that its preceding PR, #4869, went with an `applicant_classifications` instead of `classifications` relationship.

app/Http/Controllers/Api/ApplicantController.php
```diff
@@ -0,0 +1,56 @@
...
+        // Then sync new classifications to applicant, which removes all associations from the intermediate table
+        // that are not in validatedClassifications.
+        $validatedClassifications = [];
+        foreach ($data['classifications'] as $classification) {
+            $validatedClassifications[$classification['id']] = [
+                'level' => $classification['level'],
+                'order' => $classification['order'],
+            ];
+        }
+        $applicant->classifications()->sync($validatedClassifications, true);
```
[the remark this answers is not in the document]
yonikid15: The update method got refactored since we changed the default many-to-many relationship with the intermediate table, to instead have a separate model that represents the intermediate table: `ApplicantClassification.php`.

tests/Feature/ApplicantControllerTest.php
```diff
@@ -0,0 +1,140 @@
...
+
+        $expectedData = [
+            'citizenship_declaration_id' => $updateData['citizenship_declaration_id'],
+            'veteran_status_id' => $updateData['veteran_status_id'],
+            'applicant_classifications' => $applicant->applicant_classifications->toArray(),
+        ];
+        $response->assertJsonFragment($expectedData);
+        $this->assertDatabaseHas(
+            'applicant_classifications',
+            $applicant->applicant_classifications->first()->toArray()
```
tristan-orourke, line 81: This should fail as is, I think, as I'd expect that the `applicant_classification` in the database has a different id from what we sent to api.

tests/Feature/ApplicantControllerTest.php
```diff
@@ -0,0 +1,140 @@
...
+
+        // Ensure that submitting applicant classifications deletes elements that were not resubmitted
+        // and creates new elements/updates old elements.
+        $oldApplicantClassification = $applicant->applicant_classifications->first()->toArray();
+        $updateData = [
+            'citizenship_declaration_id' => CitizenshipDeclaration::inRandomOrder()->first()->id,
+            'veteran_status_id' => VeteranStatus::inRandomOrder()->first()->id,
+            'applicant_classifications' => [
+                0 => $oldApplicantClassification,
+                1 => $newApplicantClassification,
```
tristan-orourke, line 92: In this case, the two classifications will be identical. Couldn't that be a problem?

tristan-orourke: I guess, maybe not, right now? I wonder if we _should_ add some validation or database constraint stopping an applicant from adding the same classification-level combination multiple times.

tests/Feature/ApplicantControllerTest.php
```diff
@@ -0,0 +1,140 @@
...
+        ];
+        $response = $this->actingAs($applicant->user)
+            ->putJson("$this->baseUrl/applicant/$applicant->id/profile", $updateData);
+        $response->assertOk();
+
+        $applicant->refresh();
+        $expectedData = [
+            'citizenship_declaration_id' => $updateData['citizenship_declaration_id'],
+            'veteran_status_id' => $updateData['veteran_status_id'],
+            'applicant_classifications' => $applicant->applicant_classifications->toArray(),
```
tristan-orourke, line 103: Here (and a few lines later in the for loop) data is taken from `$applicant` and used in assertions. That proves that the response is correct, but it doesn't prove that `$applicant` was updated in the way we want. I think we should take the data from `$updateData` instead, like for the other fields.

tests/Feature/ApplicantControllerTest.php
```diff
@@ -0,0 +1,140 @@
...
+
+        $applicant->refresh();
+        $expectedData = [
+            'citizenship_declaration_id' => $updateData['citizenship_declaration_id'],
+            'veteran_status_id' => $updateData['veteran_status_id'],
+            'applicant_classifications' => [],
+        ];
+        $response->assertJsonFragment($expectedData);
+        $this->assertEmpty($applicant->applicant_classifications);
+        foreach ($applicant->applicant_classifications->toArray() as $applicantClassification) {
```
tristan-orourke, line 133: `$applicant->applicant_classifications->toArray()` should be an empty array at this point, so the assertDatabaseMissing test will never run.

app/Http/Controllers/Api/ApplicantController.php
```diff
@@ -0,0 +1,77 @@
...
+                if (!$newApplicantClassification) {
+                    $applicantClassification->delete();
+                }
+            }
+
+            // Update old applicant classifications and/or create them if it doesn't exist.
+            $newApplicantClassifications->map(function ($applicantClassification) use ($applicant) {
+                $newApplicantClassification = ApplicantClassification::firstOrNew(
+                    ['id' => $applicantClassification['id']]
+                );
```
tristan-orourke, line 59: I don't think we can use firstOrNew with the 'id' field. In the case of a truly new entity, we want the created object to _ignore_ the id sent in the request, and use an auto-incremented id.

tristan-orourke (changes requested):
Found a few more things, particularly some things I think are missing from the tests.

tests/Feature/ApplicantControllerTest.php
```diff
@@ -0,0 +1,140 @@
...
+    protected function setUp(): void
+    {
+        parent::setUp();
+        $this->baseUrl = 'api/v1';
+        $this->faker = \Faker\Factory::create();
+    }
+
+    public function testGetProfile(): void
+    {
+        $applicant = factory(Applicant::class)->create();
```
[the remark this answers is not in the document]
yonikid15: I added in applicant_classifications in the User factories file in PR #4869 : https://github.com/GCTC-NTGC/TalentCloud/pull/4869/commits/d6397005c67961fe0b68b7fe96f33c885e09728e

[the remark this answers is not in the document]
tristan-orourke: Thanks, yep, I missed that.

tests/Feature/ApplicantControllerTest.php
yonikid15: Oh ya, me and matt talked about this problem. Yes I agree, I'm thinking we could add some validation on both the back and front end.

app/Http/Controllers/Api/ApplicantController.php
yonikid15: I tested this out and logged the outputs, and it seems to be working as intended. If it's a new entity, it will use an auto-incremented id, and ignore the id sent in the request.

tests/Feature/ApplicantControllerTest.php
yonikid15: Turns adding the classification-level combo attributes to the `ApplicantClassification::firstOrNew` method, will filter out any duplicates.

app/Http/Controllers/Api/ApplicantController.php
```diff
@@ -0,0 +1,81 @@
...
+
+            // Update old applicant classifications and/or create them if it doesn't exist.
+            // This should remove any duplicate classification-level combinations as well.
+            $newApplicantClassifications->map(function ($applicantClassification) {
+                $newApplicantClassification = ApplicantClassification::firstOrNew(
+                    [
+                        'classification_id' => $applicantClassification['classification_id'],
+                        'level' => $applicantClassification['level']
+                    ]
+                );
```
tristan-orourke, line 64: Checking against `id` when deleting, and against `classification_id` and `level` when creating/updating can create some discrepancies. I wasn't a hundred percent sure about this, but I added a test (see 747d8b5863ad67d64273857ac45df3d023f91701) to confirm.

If you re-submit an existing application_classification (including with existing id) but with classification_id or level changed, then this _won't_ delete the previous version (correct) but it will create a _new_ object with `firstOrNew` resulting in the old and new versions both existing in the DB.

While `firstOrNew` seems convenient here, I think it would be more reliable to just check if the id exists, manually create a new object if it doesn't, and then update the values either way.

tristan-orourke (changes requested):

app/Http/Controllers/Api/ApplicantController.php
yonikid15: Just pushed up a fix now

app/Http/Controllers/Api/ApplicantController.php
```diff
@@ -0,0 +1,91 @@
...
+                    $oldApplicantClassification->classification_id = $newApplicantClassification['classification_id'];
+                    $oldApplicantClassification->fill($newApplicantClassification);
+                    $oldApplicantClassification->save();
+                } else {
+                    $applicantClassification = new ApplicantClassification();
+                    $applicantClassification->applicant_id = $newApplicantClassification['applicant_id'];
+                    $applicantClassification->classification_id = $newApplicantClassification['classification_id'];
+                    $applicantClassification->fill($newApplicantClassification);
+                    $applicantClassification->save();
+                }
```
tristan-orourke, line 78: You can save a few lines of repeated code with
```
if (!$oldApplicantClassification) {
    $oldApplicantClassification = new ApplicantClassification();
}
$oldApplicantClassification->applicant_id = $newApplicantClassification['applicant_id'];
$oldApplicantClassification->classification_id = $newApplicantClassification['classification_id'];
$oldApplicantClassification->fill($newApplicantClassification);
$oldApplicantClassification->save();
```

tristan-orourke (approved):

PR #4927, opened 2020-12-11
Task/application profile react component

### LINKED ISSUE #4853: Task - Applicant Profile - Build react component for Applicant Profile form
# Description
Build a react component for the new basic Applicant Profile form. This needs to include fields for:
- citizenship status
- veteran status
- is GC employee or not
- If a GC employee, a list of classifications

For now, simply build this UI in Storybook.

# Images
[Applicant Profile wireframe](https://xd.adobe.com/view/f66135de-6ab1-449d-ac7c-98ade3ce1fd2-a332/screen/22903d0a-6389-4e0b-a349-94d757852618)
Note: the sidebar does not need to written in React.

jaysonmc:
Resolves #4853.

### Notes:
- Since we are moving to H2, which requires creating new form components, I decided to try out the [react-form-hook ](https://react-hook-form.com/)package which is more performant than Formik (eg. less re-renders, faster mounting). 
- In the form, it asks for a current classification and any previous classifications. Since we are only storing a list of `applicant_classifications` on the applicant model, I assumed that the `applicant_classification` with `order === 0` would be the current classification.
- The form can be viewed under Applicant Profile/Basic Information in storybook.


resources/assets/js/components/ApplicantProfile/ProfileBasicInformation.tsx
```diff
@@ -0,0 +1,219 @@
+import React, { FunctionComponent } from "react";
+import { FormattedMessage, useIntl } from "react-intl";
+import { GocClassification } from "../../models/types";
+import { myBasicInformationMessages } from "../Application/applicationMessages";
+
+export interface ProfileBasicInformationProps {
+  gocClassifications : GocClassification[]
```
tristan-orourke: Looks like you have different code formatting settings than the rest of the team. I'll add some notes to the repo on our js code formatting and send you a link.

resources/assets/js/components/ApplicantProfile/ProfileBasicInformation.tsx
```diff
@@ -0,0 +1,219 @@
...
+export interface ClassificationDropdownsProps {
+  gocClassifications : GocClassification[]
+}
+
+const ClassificationDropdowns: FunctionComponent<ClassificationDropdownsProps> = ({
+  gocClassifications
+}) => {
+  const intl = useIntl();
+
+  const handleSelectedClassification : any = function(e : any){
```
tristan-orourke: We try to keep things well typed, avoiding use of `any` wherever we can.

resources/assets/js/components/ApplicantProfile/ProfileBasicInformation.tsx
```diff
@@ -0,0 +1,219 @@
...
+    this.setState({selectedClassification:e.target.value});
```
tristan-orourke: We use functional React components instead of class components [[1]](https://www.freecodecamp.org/news/functional-components-vs-class-components-in-react/). That means we use [React Hooks](https://reactjs.org/docs/hooks-intro.html) instead of `this.state` to store state and hook into the component lifecyle.

resources/assets/js/components/ApplicantProfile/ProfileBasicInformation.tsx
```diff
@@ -0,0 +1,219 @@
...
+        return
+      }
+      flag = 1
+    })
+
+    // If the flag equals one, then it is a new key value pair. Add the record to the array, as it is unique
+    if (flag == 1) {
+      uniqueClassifications.push(jsonObj)
+    }
+  })
```
tristan-orourke: This could be written a lot more succinctly. We have a number of helper functions for manipulating arrays and objects in `resources\assets\js\helpers\queries.ts`, I think some of them could be useful. If unique classifications will have different ids, then this would work:
```
const uniqueClassifications = removeDuplicatesById(
    gocClassifications.map((item) => ({
      id: item.classification.id,
      key: item.classification.key,
    })),
  );
```
If this really does need to check for unique id-key pairs, it would be nice to add a new function to `queries.ts`, similar to uniq() but that compares objects properly.

resources/assets/js/components/ApplicantProfile/ProfileBasicInformation.tsx
```diff
@@ -0,0 +1,219 @@
...
+
+  function getLevelsOfClassification(classificationKey : string) {
+    let correspondingGocClassifications : GocClassification[] = [];
+    gocClassifications.forEach(function( item : GocClassification) {
+      if (item.classification.key == classificationKey) {
+        correspondingGocClassifications.push(item)
+      }
+    })
```
tristan-orourke: This is a perfect use case for [filter](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter):
```
const correspondingGocClassifications = gocClassifications.filter(
   item => item.classification.key === classificationKey
)
```

resources/assets/js/models/types.ts
```diff
@@ -428,3 +428,17 @@ export interface Email {
...
+export interface GocClassification {
+  classification : Classification;
+  level : Number;
+  order : Number; // Used to allow user to order the list of classifications
+}
+
+export interface ProfileBasicInformation {
+  citizenship_status: CitizenshipDeclaration;
+  veteran_status: VeteranStatus;
+  current_gc_employee: Boolean;
```
tristan-orourke: Primitive types (eg number and boolean) should be lower-case.

resources/assets/js/fakeData/fakeGocClassification.ts
```diff
@@ -0,0 +1,109 @@
+import {GocClassification, Classification} from "../models/types";
+
+export const fakeGocClassifications = (
```
tristan-orourke: Why do you have a fakeGocClassification file when you're also generating fake classifications in fakeBasicInformation.ts?

resources/assets/js/components/ApplicantProfile/ProfileBasicInformation.tsx
```diff
@@ -0,0 +1,219 @@
...
+    return correspondingLevels
+  }
+
+  return (
+    <>
+      <div data-c-grid-item="base(1of3)" data-c-grid="gutter top">
+        <div data-c-grid-item="base(1of1) tl(1of2)" data-c-input="select">
+          <div>
+            <i className="fas fa-caret-down" />
+            <select  required id="SEL2" onChange={handleSelectedClassification} >
```
tristan-orourke: Looks like you're using the same id (SEL2) twice, which is not good.

resources/assets/js/components/ApplicantProfile/ProfileBasicInformation.tsx
```diff
@@ -0,0 +1,219 @@
...
+    </>
+  );
+};
+
+export const ProfileBasicInformation: React.FC<ProfileBasicInformationProps> = ({
+  gocClassifications
+}) => {
+  const intl = useIntl();
+
+  return (
```
tristan-orourke: We've been using the [Formik](https://formik.org/docs/overview) library (with [Yup](https://formik.org/docs/overview) for validation) to manage our forms in react. It helps manage all the temporary form state and handles validation. We also have a number of components in `resources\assets\js\components\Form\` to help keep the presentation of our inputs consistent.
You can do a search through our files for `Formik` to find any number of examples. `resources\assets\js\components\Application\BasicInfo\BasicInfo.tsx` might be a good one to look at first. And I'd be happy to walk you through how we're using it.

resources/assets/js/components/ApplicantProfile/ProfileBasicInformation.tsx
```diff
@@ -0,0 +1,219 @@
...
+          </div>
+          <span>This input has an error.</span>
+        </div>
+      </div>
+    </>
+  );
+};
+
+export const ProfileBasicInformation: React.FC<ProfileBasicInformationProps> = ({
+  gocClassifications
```
tristan-orourke: You'll want to pass in a ProfileBasicInformation object as well, to initialize the form values with.

resources/assets/js/components/ApplicantProfile/ProfileBasicInformation.tsx
```diff
@@ -0,0 +1,219 @@
...
+        <div data-c-input="select">
+          <label htmlFor="SEL2">{intl.formatMessage(myBasicInformationMessages.citizenStatus)}</label>
+          <span>Required</span>
+          <div data-c-grid-item="base(1of3)">
+            <i className="fas fa-caret-down" />
+            <select required id="SEL2">
+              <option disabled selected>
+                Select a status...
+              </option>
+              <option>Canadian Citizen</option>
```
tristan-orourke: For lookup tables which we don't expect to change, we've copied their values into a `lookupConstants.ts` file. You can reference CitizenStatus and VeteranStatus ids from there. Actually, it would be great if you could add entries to the `localizedConstants.tsx` file which map those ids to localized messages.
The <option> elements should use the id as their value. If you use the `resources\assets\js\components\Form\SelectInput.tsx` component, you can pass the status ids and their localized text in as a list of {value, label} objects.

resources/assets/js/components/ApplicantProfile/ProfileBasicInformation.tsx
tristan-orourke: But also, you should use the `resources\assets\js\components\Form\SelectInput.tsx` component anyway, as I describe in a later comment.

resources/assets/js/components/ApplicantProfile/ProfileBasicInformation.tsx
```diff
@@ -0,0 +1,219 @@
...
+            <label htmlFor="rB2">
+              <input id="rB2" required name="radioB" type="radio" />
+              <span>Not a GC employee</span>
+            </label>
+          </div>
+          <span>{intl.formatMessage(myBasicInformationMessages.inputEreror)}</span>
+        </div>
+
+        <label htmlFor="SEL2">Current classification and level</label>
+        <ClassificationDropdowns gocClassifications={gocClassifications} />
```
tristan-orourke: This whole Classifications section should be visible (or rendered) when user has "Current GC employee" selected.

resources/assets/js/components/ApplicantProfile/ProfileBasicInformation.tsx
```diff
@@ -0,0 +1,219 @@
...
+          <ol>
+            <li>
+              {/* Turn this into a reuseable component */}
+              <div data-c-grid="gutter top">
+                <ClassificationDropdowns gocClassifications={gocClassifications} data-c-grid-item="base(2of2) tl(2of3)" />
+                <div
+                  data-c-grid-item="base(1of1) tl(1of3)"
+                  data-c-input="select"
+                >
+                  <a href="#" data-c-color="c1">
```
tristan-orourke: Avoid using anchor elements for something that should be a button, regardless of how the mockup looks 😄 
For deleting an element from a list, we usually use a trash icon:
```
<button>
   <i className="fa fa-trash" />
</button>

File: resources/assets/js/components/ApplicantProfile/ProfileBasicInformation.tsx [hunk 14]
```diff
@@ -0,0 +1,219 @@
...
+                    Remove
+                  </a>
+                  <span>{intl.formatMessage(myBasicInformationMessages.inputEreror)}</span>
+                </div>
+              </div>
+            </li>
+          </ol>
+        </div>
+        <a href="#">{intl.formatMessage(myBasicInformationMessages.addClassification)}</a>
```
tristan-orourke: As above, this should probably be a <button>. I'm not sure how it should be styled, but [here's the guide](https://designwithclone.ca/components/button/) on how to use Clone, @bitliege's design system, to style a button in different ways.

resources/assets/js/components/ApplicantProfile/ProfileBasicInformation.tsx
```diff
@@ -0,0 +1,219 @@
...
+        <p data-c-margin="bottom(1)">
+          <FormattedMessage
+            id="profile.experience.preamble"
+            defaultMessage="This profile is also shared when you submit a job application."
+            description="First section of text on the 'My Basic Information' of the Application Timeline."
+          />
+        </p>
+        <div>
+          <p>
+            Name: <b data-c-color="c1">Gerardi Escandon</b>{" "}
```
tristan-orourke: These labels, and other english text, should also be localized. The Name should come from a ProfileBasicInformation object passed into the component.

tristan-orourke (changes requested):
Reviewing this made me realize  how many patterns/conventions we've adopted as a team that we haven't recorded anywhere. I'll have to document these things eventually. For now I've outlined a few things here, and I'm happy to chat about them in more detail.

resources/assets/js/components/ApplicantProfile/ProfileBasicInformation.tsx
```diff
@@ -0,0 +1,435 @@
...
+    const correspondingGocClassifications = gocClassifications.filter(
+      (item) => item.classification.id == classificationKey,
+    );
+
+    let correspondingLevels: string[] = [];
+    correspondingGocClassifications.forEach(function (
+      correspondingGocClassification: GocClassification,
+    ) {
+      correspondingLevels.push(correspondingGocClassification.level.toString());
+    });
```
tristan-orourke: This pattern is a good fit for using `.map` over `.forEach`

resources/assets/js/components/ApplicantProfile/ProfileBasicInformation.tsx
```diff
@@ -0,0 +1,435 @@
...
+        <div data-c-grid-item="base(1of1) tl(1of2)" data-c-input="select">
+          <div>
+            <i className="fas fa-caret-down" />
+            <select
+              defaultValue={selectedItem?.classification.id}
+              required
+              id="SEL2"
+              onChange={handleSelectedClassification}
+            >
+              <option></option>
```
tristan-orourke: We normally disable these first options with `<option value="" disabled>`, and add some text. The text is set with the `nullSelection` property if you use the `components/Select.tsx` component.

resources/assets/js/components/ApplicantProfile/ProfileBasicInformation.tsx
```diff
@@ -0,0 +1,435 @@
...
+    if (basicInformation.current_classification && basicInformation.previous_classifications.length > 0) {
+      return currentEmployeeIdEnum.Yes;
+    }
+    else if (basicInformation.current_classification && basicInformation.previous_classifications.length == 0) {
+      return currentEmployeeIdEnum.Previous;
+    }
+    else {
+      return currentEmployeeIdEnum.No;
+    }
+  };
```
tristan-orourke: Do you need to calculate this value? The way your props are set up, isn't it directly included in basicInformation?

resources/assets/js/components/ApplicantProfile/ProfileBasicInformation.tsx
```diff
@@ -0,0 +1,435 @@
...
+    return correspondingLevels;
+  }
+
+  return (
+    <>
+      <div data-c-grid-item="base(1of3)" data-c-grid="gutter top">
+        <div data-c-grid-item="base(1of1) tl(1of2)" data-c-input="select">
+          <div>
+            <i className="fas fa-caret-down" />
+            <select
```
tristan-orourke: We want to wrap this select input in a Formik component so that we can manage all initialization, validation, and submission of all our form values in the same place (the root Formik component). Typically we'd use a `<FastField>` to wrap the  `Form/SelectInput` component like so: `<FastField component={SelectInput} ...otherProps>`.

Its a bit more complicated using Formik with a list of items that can be added or removed like this, but we've done it once before. See `JobBuilder/Tasks/JobTasks.tsx` where a `<FieldArray>` wraps a render function that renders each item.

resources/assets/js/stories/ApplicantProfile/ProfileBasicInformation.stories.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+const stories = storiesOf("Applicant Profile/Basic Information", module).addDecorator(
+  withIntl,
+);
+
+stories.add(
+  "Current GC Employee",
+  (): React.ReactElement => {
+    const [basicInformation, setBasicInformations] = useStorybookState(fakeBasicInformation());
+    const [gocClassifications]  = useStorybookState(fakeGocClassifications());
```
tristan-orourke: Since these values are never updated, you don't need to use the state hook. You can call the `fake___()` functions directly in the component properties below:
```
<ProfileBasicInformation
   basicInformation={fakeBasicInformation()}
   ...
/>
```

resources/assets/js/components/ApplicantProfile/ProfileBasicInformation.tsx
```diff
@@ -0,0 +1,435 @@
...
+    });
+  };
+
+  const addExperience = function () {
+    // If the user has not populated their most recent experience, prevent adding a second empty row
+    if (
+      previousExperience.filter((experience) => {
+        return experience.classification.id == 0;
+      }).length > 0
+    ) {
```
tristan-orourke: I'm not sure how, but I think there might be a bug in this logic? When playing with this in Storybook, it sometimes doesn't create a row when I expect it to.
Also, it seems to me it might be nicer UX to disable the button when we don't want it to do anything, rather than abort silently.

tristan-orourke (changes requested):
Seems like you don't have the linter fully set up yet. I'm getting a lot of linting errors in ProfileBasicInformation.tsx. Mostly minor stuff, but some of it useful.

The other issue is that the Classification select inputs aren't tied into Formik. You can reference `resources\assets\js\components\JobBuilder\Tasks\JobTasks.tsx` for an example of how we've used Formik's <FieldArray> to wrap an expandable & deletable list of items. You can also view it in Storybook, under the Job Poster Builder section.

tristan-orourke:
I forgot to link this before, but I added some notes on our linting setup to the wiki:
https://github.com/GCTC-NTGC/TalentCloud/wiki/Coding-Style-Guide#typescript-standards

resources/assets/js/components/ApplicantProfile/ProfileBasicInformation.tsx
jaysonmc: Good call. You are correct.

resources/assets/js/components/ApplicantProfile/BasicInfo/ProfileBasicInfo.tsx
```diff
@@ -0,0 +1,495 @@
...
+        {intl.formatMessage(messages.personalEmail)}:{" "}
+        <span data-h2-font-color="b(theme-1)" data-h2-font-weight="b(700)">
+          {email}
+        </span>
+      </p>
+      <p data-h2-margin="b(bottom, 1)">
+        {intl.formatMessage(messages.toChangeGoTo)}:{" "}
+        <a
+          data-h2-font-color="b(theme-1)"
+          href={getApplicantProfileEndpoint(applicantId)}
```
tristan-orourke: You have the api endpoint here :P

resources/assets/js/components/ApplicantProfile/BasicInfo/ProfileBasicInfo.tsx
```diff
@@ -0,0 +1,495 @@
...
+    id: 0,
+    applicant_id: applicantId,
+    classification_id: -1,
+    level: -1,
+    order: -1,
+  };
+
+  const [currentEmployeeStatus, setCurrentEmployeeStatus] = useState<
+    number | null
+  >(gcEmployeeStatus);
```
tristan-orourke: Hook Form provides the `watch` function for cases like this. See this example: https://codesandbox.io/s/13ykqx4wx7

resources/assets/js/components/ApplicantProfile/BasicInfo/ProfileBasicInfo.tsx
```diff
@@ -0,0 +1,495 @@
...
+    defaultValues,
+    resolver: yupResolver(validationShema),
+  });
+
+  // Custom hook for working with uncontrolled field arrays (dynamic inputs).
+  // https://react-hook-form.com/api#useFieldArray
+  const { fields, append, remove } = useFieldArray({
+    control,
+    name: "previousClassifications",
+    keyName: "previousClassifications",
```
tristan-orourke: I think the purpose of this keyName field is to rename the unique id that useFieldArray generates for each entry. We do want to rename it, because our ApplicantClassification object already has an id field, but we need to make sure to use it later. I would rename it to simply "key", because its used as the ReactElement key like so:
```
fields.map((previousClassification, index) => (
   <li key={previousClassification.key}></li>
))
```
Currently the page generates `Encountered two children with the same key` console errors, I think this will fix it.

resources/assets/js/components/ApplicantProfile/BasicInfo/ProfileBasicInfo.tsx
```diff
@@ -0,0 +1,495 @@
...
+                >
+                  {localizeFieldNonNull(locale, classification, "name")}
+                </Select.Option>
+              ))}
+            </Select>
+            <Select
+              name="currentLevel"
+              defaultValue={`${currentLevel}`}
+              required
+              register={register}
```
tristan-orourke: The notes on [useFieldArray](https://react-hook-form.com/api#useFieldArray) mention that register must be called **as a function** within the map function: `ref={register()}`. This ensures the registers are updated correctly when fields are added or removed.

resources/assets/js/components/H2Components/Form.tsx
```diff
@@ -0,0 +1,33 @@
...
+const Form: React.FunctionComponent<FormProps> = ({
+  onSubmit,
+  children,
+  ...rest
+}) => {
+  React.useEffect((): void => {
+    h2FormIndicateRequiredFields("all");
+    h2FormValidationCheck("all");
+    h2FormAccordionToggle("all");
+    h2FormPasswordToggle("all");
```
tristan-orourke, line 23: I want to look into these a little deeper. I'm a little worried about something conflicting with Yup/React validation.

resources/assets/js/components/H2Components/Select.tsx
```diff
@@ -0,0 +1,158 @@
+import * as React from "react";
+import { useIntl } from "react-intl";
+import { FieldName, FieldValues, RegisterOptions } from "react-hook-form";
```
tristan-orourke: Unused imports

resources/assets/js/components/H2Components/Select.tsx
```diff
@@ -0,0 +1,158 @@
...
+const Option: React.FunctionComponent<OptionProps> = ({ value, children }) => {
+  useSelectContext(); // Ensures sub-component can only be used within the Option component.
+  return <option value={value}>{children}</option>;
+};
+
+interface SelectComposition {
+  Option: React.FunctionComponent<OptionProps>;
+}
+
+const Select: React.FunctionComponent<SelectContext> & SelectComposition = (
```
tristan-orourke, line 60: This seems more specific to Talent Cloud than the other components in this folder (especially with use of react-intl and particular messages). I don't have anything to say about the component, but I think it should be in a different folder.

resources/assets/js/components/ApplicantProfile/BasicInfo/ProfileBasicInfo.tsx
```diff
@@ -0,0 +1,495 @@
...
+  name: string;
+  previousClassifications: ApplicantClassification[];
+  veteranStatus: number | null;
+  handleUpdateProfile: (data: ApplicantProfile) => Promise<void>;
+}
+
+export const ProfileBasicInfo: React.FC<ProfileBasicInfoProps> = ({
+  applicantId,
+  currentClassification,
+  currentLevel,
```
tristan-orourke: Since the backend model currently just has a single list of applicant_classifications, I'm worried splitting it into current and previous classifications here might make it difficult to communicate back and forth.
What do you think of just passing one list of classifications, and treating the zeroth index as current? If currentEmployStatus === "yes" then"currentClassifications" could just be index 0, `previousClassifications[0]`, and we could skip index 0 within the `fields.map` function.

tristan-orourke (changes requested):
There are currently some console errors caused by duplicate react keys, I think they have to do with the keyName field of useFIeldArray.
There's also a bug where if you remove one classification field, **no** past classifications are submitted. I think that will be fixed by calling register like a function: `ref={register()}`

resources/assets/js/components/ApplicantProfile/BasicInfo/ProfileBasicInfo.tsx
```diff
@@ -0,0 +1,495 @@
...
+          .typeError(intl.formatMessage(validationMessages.required))
+          .required(intl.formatMessage(validationMessages.required)),
+        level: Yup.number()
+          .min(1, intl.formatMessage(validationMessages.invalidSelection))
+          .max(9, intl.formatMessage(validationMessages.invalidSelection))
+          .typeError(intl.formatMessage(validationMessages.required))
+          .required(intl.formatMessage(validationMessages.required)),
+      }),
+    ),
+  });
```
tristan-orourke: I know this is how we've done react form validation up until now, and I'm very happy to know that our existing Yup validation objects could be ported directly to Hook Form, but I'm thinking now that this is overkill.
We can control a field being required at the level of HTML, and its probably unnecessary to control for valid values in a Select input. The user would have to edit the values of the fields in the inspector to trigger those errors, at which point we're relying on backend validation anyway.

tristan-orourke: Which means **none** of this Yup validation might be necessary, actually 🤔. (In this case anyway.)

resources/assets/js/components/H2Components/Form.tsx
tristan-orourke: I've looked at the scripts - I do think there's going to be a problem if we do more complex yup validation (eg word counts) but I'd say that's more of a H2 discussion and outside the scope of this PR. I'll accept this as is for now.

resources/assets/js/components/ApplicantProfile/BasicInfo/ProfileBasicInfo.tsx
yonikid15: It says to call is as a function when there are no validation rules. But since we are removing the unnecessary validation you mentioned below this will called as a func now

resources/assets/js/components/H2Components/Select.tsx
yonikid15: All the messages are the default H2 messages that shouldn't be changed ever.

resources/assets/js/components/ApplicantProfile/BasicInfo/ProfileBasicInfo.tsx
yonikid15: There's supposed to be a values to forms function to match the correct structure before passing it to the backend but I like this method better too.

resources/assets/js/components/H2Components/Select.tsx
tristan-orourke: Right, but if we want to build components that can be included in the H2 library, they shouldn't use `react-intl` at all. It doesn't matter for this PR though, and we should probably have another conversation about whether that's still the goal.

resources/assets/js/components/ApplicantProfile/BasicInfo/ProfileBasicInfo.tsx
```diff
@@ -0,0 +1,499 @@
...
+  classifications,
+  email,
+  gcEmployeeStatus,
+  name,
+  veteranStatus,
+  handleUpdateProfile,
+}) => {
+  const intl = useIntl();
+  const locale = getLocale(intl.locale);
+  const currentClassification = applicantClassifications.shift(); // Removes first element from array;
```
tristan-orourke: Modifying an object directly which was passed in as a prop? That's bad practice :finger_wag: At least clone it first!
I'd prefer:
```
const currentClassification = applicantClassifications[0] ?? null; // This explicitly leaves it null if applicantClassifications is an empty array.
const previousClassifications = applicantClassifications.slice(0);
```
or, if you want to be fancy,
```
const [currentClassification, ...previousClassifications] = applicantClassifications;
```
Though I'm not sure how the second version treats an empty array. 🤔 Actually, [stick with the first one](https://github.com/microsoft/TypeScript/issues/38287).

tristan-orourke (changes requested):
Just one small change needs to be made

resources/assets/js/components/ApplicantProfile/BasicInfo/ProfileBasicInfo.tsx
tristan-orourke: Okay, I've made this change.

tristan-orourke (changes requested):
Unfortunately, there's still 2 bugs I found:
1. When you're not in government and submit with 0 classifications, it currently submits undefined instead of an empty array for classifications.
2. When you're currently in government but don't have any old classifications, it fails validation (silently) and doesn't allow you to submit.

tristan-orourke:
> Unfortunately, there's still 2 bugs I found:
> 
> 1. When you're not in government and submit with 0 classifications, it currently submits undefined instead of an empty array for classifications.
> 2. When you're currently in government but don't have any old classifications, it fails validation (silently) and doesn't allow you to submit.

Fixed with 01ed54d47ea5460176f71e60ba33883a4fb2b261!

tristan-orourke (approved):

PR #4952, opened 2020-12-15
Installs H2

bitliege:
You can use H2 anywhere on the site now, as long as you wrap the code in a `div` with the `data-h2-system` attribute.

If you want to create an entirely new page, I've added an H2 layout in the views.


resources/views/layouts/base-h2.html.twig
```diff
@@ -0,0 +1,23 @@
...
+    <main>
+      {% block body %}{% endblock %}
+    </main>
+  </div>
+  {% include "common/footer" %}
+  {% include "common/scripts" %}
+  {% block scripts %}{% endblock %}
+  <script async defer
+    src="https://cdn.jsdelivr.net/npm/@hydrogen-design-system/system@latest/dist/compiled/latest/system.min.js">
+  </script>
```
tristan-orourke, line 21: This is already in the included scripts template, it doesn't need to be here as well.

tristan-orourke (changes requested):

tristan-orourke (approved):

PR #4960, opened 2020-12-16
Task - Experience step client side validation

### LINKED ISSUE #4937: Task - Timeline - Experience Step front-end validation
# Description
Clicking "Save and Continue", should only do so if all the required skills are connected to an experience. If the validation fails, then focus on the error message.

image

yonikid15:
Resolves #4937 

### Notes:
- If any required skills have not been attached to an experience, then an error alert will popup and receive focus.  
- The error alert has been placed above the experience buttons so that the user can tab to them after reading the error message.
- I tried to find most optimal approach for accessibility.


resources/assets/js/components/Application/Experience/Experience.tsx
```diff
@@ -706,14 +733,51 @@ export const ExperienceStep: React.FunctionComponent<ExperienceStepProps> = ({
...
 }) => {
   const intl = useIntl();
+  const [hasError, setHasError] = useState(false);
+
+  // Hack fix to focus on error message after it has been added to the DOM.
+  useEffect(() => {
+    if (hasError) {
+      focusOnElement("experience-step-form-error");
+    }
+  }, [hasError]);
```
tristan-orourke, line 743: So, why is this necessary when the submit button also calls `focusOnElement`?

tristan-orourke: I see, the hook triggers focus the first time hasError is set to true, because the div isn't rendered yet when the onClick event happens. On subsequent submit attempts, if the error element is still there it can be focused right in the onClick handler.

tristan-orourke (approved):

resources/assets/js/components/Application/Experience/Experience.tsx
yonikid15: So I was trying to get the alert message to focus after clicking on the save and continue button. However, the focus is triggered (in the `onClick` event method) before the element exists in the DOM. Therefore, I used the `useEffect` hook to trigger the `focusOnElement` method for that first click after react re-renders and sets `hasError` to true.

PR #5053, opened 2021-01-08
Add skill search bar UI

### LINKED ISSUE #4855: Task - Build Skill Search bar UI
# Description
Build a react component, given a list of skills, allows you to search and displays results. For this issue, build this in storybook.

For now, do simple case-insensitive string matching against the skill titles for the search.

If possibly, consider creating the search bar component in such a way that it could be reused for other purposes later (though the way the results are formatted will be specific to this Skill search).

For this task, don't worry about actually saving added skills to the server, but _do_ have the component accept the list of skills that are already saved to the applicant, and present their results accordionly.

# Images
The interactive wireframe can be found [here](https://xd.adobe.com/view/c2c5a303-1dcd-4436-bec3-eea0a3ca3523-9796/).
But note that this issue the search bar and results:
Search Bar

mnigh:
Resolves #4855


resources/assets/js/components/ApplicantProfile/Skills/SkillSearchResults.tsx
```diff
@@ -0,0 +1,199 @@
...
+
+export const SkillSearchResults: React.FunctionComponent<SkillSearchResultsProps> = ({
+  status,
+  results,
+  handleAddSkill,
+}: SkillSearchResultsProps): React.ReactElement => {
+  const intl = useIntl();
+  React.useEffect((): void => {
+    h2ComponentAccordionAddTriggerEvent();
+  }, [results]);
```
tristan-orourke, line 156: I think it would be safer to leave the second arg out entirely, and have the effect run on every re-render.

tristan-orourke: Actually, I believe this hook should be in the Item component, since if that component re-rendered, and the parent component didn't, we'd still need to re-attach the event listeners.

resources/assets/js/components/ApplicantProfile/Skills/SkillSearchResults.tsx
```diff
@@ -0,0 +1,199 @@
...
+  resultAccordionAdded: {
+    id: "results.accordion.added",
+    defaultMessage: "Added!",
+    description: "The message displayed if skill added.",
+  },
+});
+
+interface SkillSearchResultsProps {
+  status: string;
+  results: Array<Skill> | null;
```
tristan-orourke, line 37: It doesn't actually expect `Array<Skill>`, it expects `Array<Skill & {isChecked?: boolean}>`. In fact, I think you should make it not optional by removing the `?`. Right now, SkillSearchBar always passes down an array of Skills with no isChecked status, and technically that's valid (but I don't think it should be).

resources/assets/js/components/ApplicantProfile/Skills/SkillSearchResults.tsx
```diff
@@ -0,0 +1,199 @@
...
+  handleAddSkill: (skillId: number) => Promise<Skill>;
+}
+
+const SkillSearchResultItem: React.FunctionComponent<SkillSearchResultItemProps> = ({
+  item,
+  handleAddSkill,
+}: SkillSearchResultItemProps): React.ReactElement => {
+  const intl = useIntl();
+  const locale = getLocale(intl.locale);
+  const [added, setAdded] = React.useState(false);
```
tristan-orourke, line 57: I think it would be worth leaving a comment here explaining why we have an `added` state in addition to `isChecked` coming from the prop - the fact that we want to show different UI if we've just recently added it compared to sometime in the past, and this helps distinguish that.

resources/assets/js/components/ApplicantProfile/Skills/SkillSearchBar.tsx
```diff
@@ -0,0 +1,121 @@
...
+                    <span data-h2-button-label>
+                      {inputTitle ||
+                        intl.formatMessage(formMessages.searchButtonText)}
+                    </span>
+                  </button>
+                </div>
+              </div>
+            </div>
+            <SkillSearchResults
+              status={status}
```
tristan-orourke, line 110: I don't think it's necessary to pass status down to the SkillSearchResults. Instead, you can conditionally not render it at all until we've submitted a search:
```
{status === "submitted" && (
  <SkillSearchResults ... />
  )}

tristan-orourke: PS: smart use of Formik's setStatus()! I didn't even know about this function. 😄

tristan-orourke (changes requested):
Works, just a few minor comments. Oh, and I wonder if we should not make the search field `required`. Seems funny to me to have the red "Required" text when its the only field in the form. Also, someone might want to clear the results list by submitting an empty string... maybe?

resources/assets/js/components/ApplicantProfile/Skills/SkillSearchResults.tsx
mnigh: @tristan-orourke I had previously tried both leaving out the second argument as well as moving the `useEffect` hook to then Item component (with and without the second argument). All of these variations resulted in the accordion not being able to toggle its expansion. Is there something else I'm missing here?

tristan-orourke: Hmm, If that's the case, more likely I'm missing something here.

tristan-orourke (approved):

PR #5074, opened 2021-01-12
Update eslint settings and format related files

Khristophor:
### Notes:
- I'm sorry
- I added a helper method to grab the first 6 characters of a whitespace-stripped string to use as a key, there were a couple components complaining about not having a key, and the iterable is a string so I couldn't use an ID or anything useful
- The bulk of the changed files is just whitespace/line length/etc

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -6,14 +6,14 @@
         "jest": true
     },
     "extends": [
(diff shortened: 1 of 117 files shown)


yonikid15 (approved):
Looks good to me 👍 @tristan-orourke @mnigh

PR #5081, opened 2021-01-14
Task - Improve storybook build performance

yonikid15:
### Notes:
- Removes [react-docgen-typescript-loader](https://github.com/strothj/react-docgen-typescript-loader). Removing this improved performance significantly (build and hot-reload). 
- Explicitly sets the `include` key within `webpack.config.js` to only check within `/js` folder.
- Sets the `stories` value, in the storybook config (`main.ts`), to only parse within the `/js` folder.

Previous build (~1.95 min) and hot-reload(~7 sec) performance:
image

New build (25 sec) and hot-reload (~2 sec ) performance: 
image


Khristophor (approved):
Nice work! This is wayyyyy faster.

Issue #4858, opened 2020-12-01
Task - Build Applicant Profile Skills UI


tristan-orourke:
# Description
Build a React component that displays the Applicant's skills, based on the wireframe. This should be built in Storybook.

# Images
[Wireframe](https://xd.adobe.com/view/c2c5a303-1dcd-4436-bec3-eea0a3ca3523-9796/) (Click UI Design Tools to see the correct accordion format. Research methods has an alternative, unused design.)

# Acceptance criteria
- [ ] Sortable by alphabetical, or by Group (ie top-level Skill Category). Ignore the "Journeys" option in the wireframe.
- [ ] Each Skill is an expandable accordion which lists linked Experiences
  - [ ] Each accordion shows the skill definition.
  - [ ] Each experience list item shows Title and Subtitle (eg Company, for work) and looks like a link.
  - [ ] If justification is empty, show a prompt and link to My Experience profile page
  - [ ] Also show a delete button at the bottom of skill accordion
- [ ] Clicking an Experience "link" opens a modal with full details of the Experience
- [ ] An Experience details modal also contains a link which takes you to the Experience page and opens the appropriate Experience modal (see #4863)

PR #5087, opened 2021-01-18
Feature - Applicant Profile Skills UI

Khristophor:
Relates to the bottom section of [the wireframes](https://xd.adobe.com/view/c2c5a303-1dcd-4436-bec3-eea0a3ca3523-9796/) for the profile skills.

### Notes:

I pushed up WIP code, someone will have to take this over. The reordering column values (`lft`/`rgt`) on the `skill_categories` table don't seem to make any sense currently, so I wasn't able to finish up any helper functions for grouping/ordering skills for the frontend. I have a first pass on a sorting function for models with localized fields, but it's also throwing a type error that I'm not sure about.


mnigh:
Resolves #4858.

### Notes
Main list component, as well as child components, can be viewed as stories in storybook:
- Applicant Profile / Skills / List
- Applicant Profile / Skills / Accordion
- Applicant Profile / Skills / Accordion (No Experiences)
- Applicant Profile / Experience Details / Work
- Applicant Profile / Experience Details / Education
- Applicant Profile / Experience Details / Community
- Applicant Profile / Experience Details / Award
- Applicant Profile / Experience Details / Personal

resources/assets/js/components/ApplicantProfile/Skills/Accordion.tsx
```diff
@@ -0,0 +1,633 @@
...
+                    </p>
+                  </Dialog.Content>
+                  <Dialog.Actions
+                    data-h2-padding="b(tb, 1) b(rl, 2)"
+                    data-h2-align="b(center)"
+                    data-h2-grid="b(middle, expanded, flush, 1)"
+                  >
+                    <div data-h2-grid-item="b(1of2)" data-h2-align="b(left)">
+                      <div data-h2-grid-content>
+                        <Dialog.ActionBtn
```
tristan-orourke, line 557: Needs an onClick. The SkillAccordion should probably accept a callback prop, something like `deleteSkill: (skillId: number) => Promise<void>`, which can be called when you click this.

resources/assets/js/components/ApplicantProfile/Skills/Accordion.tsx
```diff
@@ -0,0 +1,633 @@
...
+                      >
+                        {intl.formatMessage(
+                          messages.missingExperiencesLinkText,
+                        )}
+                      </a>
+                    ),
+                  })}
+                </p>
+                <p data-h2-padding="b(tb, .5)">
+                  <button
```
tristan-orourke, line 609: This also need the onClick deleteSkill.

resources/assets/js/components/ApplicantProfile/Skills/List.tsx
```diff
@@ -0,0 +1,202 @@
...
+  skillCategories,
+  skillSkillCategories,
+  skills,
+  applicantId,
+}) => {
+  const intl = useIntl();
+  const locale = getLocale(intl.locale);
+  const [sortType, setSortType] = useState<SortTypes>(SortTypes.Group);
+
+  const skillCategoriesGroupedObject = skillCategories
```
tristan-orourke, line 84: This looks like a lot of iterating, it might be worth wrapping this in `useMemo`.

tristan-orourke (changes requested):
Nice work, looks great, works smoothly! Just needs a deleteSkill handler passed in and attached to the buttons.

tristan-orourke:
@mnigh @bitliege I noticed a weird dialog bug in these stories. It probably has nothing to do with this PR but we should look into at some point. 
Some of the dialogs close correctly when I press escape, but for some of them the overlay disappears and the dialog remains until I click a close button. I'm not sure what the difference is.

resources/assets/js/components/ApplicantProfile/Skills/List.tsx
```diff
@@ -0,0 +1,202 @@
...
+          } else {
+            accumulator[currentCategory.parent_id] = {
+              title: titleValue,
+              skills: skillsValue,
+            }; // Add child category to new parent category group.
+          }
+        }
+        return accumulator;
+      },
+      [],
```
tristan-orourke, line 126: With the way parent_id is used as a key, and accordion to the type attributed to accumulator, this should be an empty object `{}` not an array. I'm surprised TypeScript didn't warn about that 🤔

resources/assets/js/components/ApplicantProfile/Skills/List.tsx
```diff
@@ -0,0 +1,202 @@
...
+    );
+
+  const skillCategoriesGrouped = Object.keys(skillCategoriesGroupedObject).map(
```
tristan-orourke, line 133: I think you want filter instead of map here. Having entries of `false` in the resulting array isn't useful.

resources/assets/js/components/ApplicantProfile/Skills/List.tsx
```diff
@@ -0,0 +1,212 @@
...
+        />{" "}
+        <SortButton
+          text={intl.formatMessage(messages[SortTypes.Alpha])}
+          active={sortType === SortTypes.Alpha}
+          handleClick={() => setSortType(SortTypes.Alpha)}
+        />
+      </p>
+      {sortType === "group" &&
+        skillCategoriesGrouped.map((category) => (
+          <div key={category.catId}>
```
tristan-orourke, line 167: `catId` is actually undefined. It was never a property attached to elements, it was the key of that element (or index, when its an array). You could get the index like so:
```
skillCategoriesGrouped.map((category, i) => (
      <div key={i}>
```
But I don't think that is the best key to use.

tristan-orourke (approved):

PR #5099, opened 2021-01-21
Task - Applicant - Explore skills UI

### LINKED ISSUE #4854: Task - Build Explore Skills UI
# Description
Build a React component for the Explore Skills UI. For now, build this in Storybook.

Given the full list of Skills, and the full list of SkillCategories, use the root Categories to construct the "Explore Skills" buttons, and their child categories (and linked skills) to construct the multi-tiered menu modals.

Note: styling and ensuring we meet accessibility requirements may require some discussion with @bitliege and @gggrant.

# Images
The interactive wireframe can be found [here](https://xd.adobe.com/view/c2c5a303-1dcd-4436-bec3-eea0a3ca3523-9796/).

yonikid15:
Resolves #4854 

### Notes:
- The FindSkillsModal can be viewed on storybook under Applicant Profile -> Find Skills Modal.
- Still have a few things left to finish:
  - [x] Remove hardcoded text and replace with formatted messages
  - [x] Break down FindSkillsModal into smaller components


resources/assets/js/components/FindSkillsModal.tsx
```diff
@@ -0,0 +1,481 @@
...
+                      data-h2-border={`${
+                        index + 1 !== parentSkillCategories.length
+                          ? "b(gray-2, bottom, solid, thin)"
+                          : ""
+                      }`}
+                      data-h2-margin="b(tb, 0)"
+                    >
+                      <Accordion triggerPos="left">
+                        <Accordion.Btn
+                          buttonStyling=""
```
tristan-orourke, line 148: Redundant attribute?

resources/assets/js/components/FindSkillsModal.tsx
```diff
@@ -0,0 +1,481 @@
...
+                          <ul
+                            data-h2-padding="b(all, 0)"
+                            className="no-list-style-type"
+                          >
+                            {childrenSkillCategories.map(
+                              (childSkillCatergory) => {
+                                return (
+                                  <li key={childSkillCatergory.key}>
+                                    <div
+                                      data-h2-dialog-actions
```
tristan-orourke, line 192: This comment isn't a critique, but I'm just wondering what this `data-h2-dialog-actions` attribute does. Is it just for styling?

resources/assets/js/components/FindSkillsModal.tsx
```diff
@@ -0,0 +1,481 @@
...
+          data-h2-margin="b(all, 0)"
+          data-h2-bg-color="b(gray-1, 1)"
+        >
+          <div data-h2-align="b(left)" data-h2-grid-item="b(1of2)">
+            <Dialog.ActionBtn
+              buttonStyling="stop, round, solid"
+              data-h2-padding="b(rl, 2) b(tb, .5)"
+              data-h2-bg-color="b(white, 1)"
+              onClick={() => {
+                setSkillsResults([]);
```
tristan-orourke, line 458: I'm having some trouble testing the cancel button, but I think this should do all the same things as the Back button?

resources/assets/js/components/FindSkillsModal.tsx
```diff
@@ -0,0 +1,481 @@
...
+                          ) : (
+                            <button
+                              data-h2-button=""
+                              data-h2-grid-item="b(1of4)"
+                              type="button"
+                              disabled={isOldSkill}
+                              onClick={() => {
+                                // If the skill has been selected then remove it.
+                                // Else, if the has not been selected then add it to addedSkills list.
+                                if (isAdded) {
```
tristan-orourke, line 406: There's very similar logic when `setExpandedAccordions` is used, I wonder if a little helper function would be nice, that removes an element from an array its already present, and removes it if it's not.

resources/assets/js/components/FindSkillsModal.tsx
```diff
@@ -0,0 +1,481 @@
...
+                                                  locale
+                                                ],
+                                              description:
+                                                childSkillCatergory.description[
+                                                  locale
+                                                ],
+                                            });
+                                            setSkillsResults(
+                                              childSkillCatergory.skills,
+                                            );
```
tristan-orourke, line 220: Right now we're setting all these state hooks at the same time, and then clearing them all at the same time when the back button is clicked. It makes me think there might be some redundancy here. Like it might be simpler, logic-wise, to just use one state hook and then compute the other values, something like:
```
setSelectedCategory(childSkillCategory)
resultsSectionText.name = {name: selectedCategory.name[locale], ...}
skillResults = selectedCategory.skills
firstVisit = selectedCategory === null
```

tristan-orourke (changes requested):
Really great job on this, it looks great and there's some complicated logic here that's very easy to read and understand. I have a few suggestions that might simplify that slightly.

Unfortunately, I'm seeing some console errors and warnings when I close the modal, and I'm not able to re-open it after that. I think its more of an issue with the H2 Dialog than this story, but I want to understand that issue before approving.

resources/assets/js/components/FindSkillsModal.tsx
```diff
@@ -0,0 +1,481 @@
...
+                            setExpandedAccordions(
+                              expandedAccordions.includes(key)
+                                ? expandedAccordions.filter(
+                                    (accordionKey) => accordionKey !== key,
+                                  )
+                                : [...expandedAccordions, key],
+                            )
+                          }
+                        >
+                          <p data-h2-font-weight="b(700)">{name[locale]}</p>
```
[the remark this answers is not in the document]
yonikid15: Yes I can, I knew something was off

resources/assets/js/components/FindSkillsModal.tsx
yonikid15: Nope, no special styling. I think I was just copy/pasting code and forgot to remove that.

resources/assets/js/components/FindSkillsModal.tsx
yonikid15: This was addressed in the hydrogen slack channel. It's the same issue we had with the accordion event listener. It should be resolved in the next release.

resources/assets/js/components/FindSkillsModal.tsx
yonikid15: That's an excellent idea. so would the other values just be props or variables within the component?

tristan-orourke: The other values would be constants defined somewhere below where the SelectedCategory state hook is defined, and above the return statement.

resources/assets/js/helpers/queries.ts
```diff
@@ -236,3 +236,15 @@ export function removeDuplicatesById<T extends { id: number }>(
...
   return num <= 0 ? 0 : num - 1;
 }
+
+/**
+ * If the element already exists in the array, then remove it, and return array.
+ * Otherwise, if the element does not exist in the array, add it to the array, and return new array.
+ * @param element
+ * @param array
+ */
+export function createOrRemove<T>(element: T, array: T[]): T[] {
```
tristan-orourke, line 246: I would name it slightly differently, like addOrRemove, but its fine 😛

resources/assets/js/components/FindSkillsModal.tsx
```diff
@@ -81,15 +85,21 @@ const FindSkillsModal: React.FunctionComponent<FindSkillsModalProps> = ({
       skillCategory.depth === 1 && skillCategory.parent_id === 0,
   );
 
+  // List of new Skills that will be saved to the user on submit.
```
tristan-orourke, line 89: With so many state hooks here, comments are a great addition 👍

tristan-orourke (approved):

PR #5132, opened 2021-01-27
Remove ClassificationId constant from frontend

### LINKED ISSUE #4892: Task - Remove ClassificationId constant from frontend
# Description
#3908 started the process of stripping the Classification constants from the frontend, but because we have multiple areas at various stages of hook/redux usage it becomes difficult to simply strip out this functionality.

Areas affected:
- Application v2
- Job Poster Builder
- Storybook stories
- HR Index
- Review Application

The hook is available in `hooks/jobBuilderHooks.ts` as `useLoadClassifications()` (we should probably move it out of here at it's more global), and I _think_ just needs to be called in the relevant page files and passed as a Classification array prop, but taking out the `ClassificationId` constant also affects other places like `models/jobUtil.ts` since we have a helper function for constructing a classification string from a key and level concatenation. With the new `name` field added in #3908 this shouldn't be necessary.

mnigh:
Resolves #4892.


yonikid15 (changes requested):
Looks good so far, one more thing left though. There's still a few places using the [`educationMessages`](https://github.com/GCTC-NTGC/TalentCloud/blob/task/remove-classificationid-constant/resources/assets/js/components/JobBuilder/Details/JobDetailsMessages.tsx#L148), in `JobDetails.tsx `, and `EducationSubForm.tsx`. We should be grabbing the education requirements from the classifications in these components as well.

yonikid15 (approved):
Looks good! However, I found a unrelated issue on the Job Details page. The page finishes loading before the classifications are done being fetched, causing a delay before the select box has any options. I'll make a separate issue for this.

PR #5136, opened 2021-01-28
Task - Hydrogen react components (Accordion, Dialog, Button)

yonikid15:
### Notes:
- The components can be viewed in the H2 Components section in storybook.
- Uses the compound component pattern. This pattern allows a component and it's sub-components to share an implicit state, and handle logic together (using the `useContext` hook). This pattern should help create more declarative and flexible components.


tristan-orourke:
> using the useContext hook
👏 Yay! I've been wanting to get used to using this more!

yonikid15:
> > using the useContext hook
> > 👏 Yay! I've been wanting to get used to using this more!

Ya same 😄. Using the compound component pattern with our hydrogen components will make building the larger components much easier to customize and easier to read through.

resources/assets/js/components/H2Components/Dialog.tsx
```diff
@@ -0,0 +1,184 @@
...
+const useDialog = (): Partial<DialogContext> => {
+  const context = React.useContext(DialogContext);
+  if (!context) {
+    throw new Error("This component must be used within a <Dialog> component.");
+  }
+  return context;
+};
+
+const Actions: React.FunctionComponent = (props) => {
+  const { id = "dialog" } = useDialog();
```
tristan-orourke, line 32: This isn't used

resources/assets/js/components/H2Components/Dialog.tsx
```diff
@@ -0,0 +1,184 @@
...
+  useDialog();
+  const intl = useIntl();
+  const { children } = props;
+  return (
+    <Button data-h2-dialog-exit-trigger data-h2-bg-color="b(white, 0)">
+      <i className="fas fa-times" aria-hidden="true" />
+      <span data-h2-visibility="hidden">
+        {intl.formatMessage(messages.closeDialog)}
+      </span>
+    </Button>
```
tristan-orourke, line 81: This component seems much less general than the others, and doesn't actually use `children` anywhere. Instead of including text, shouldn't that come from `children`?

resources/assets/js/components/H2Components/Dialog.tsx
```diff
@@ -0,0 +1,184 @@
...
+
+interface HeaderProps {
+  className?: string;
+}
+
+const Header: React.FunctionComponent<HeaderProps> = (props) => {
+  useDialog();
+  const { className, children } = props;
+  return (
+    <div data-h2-dialog-title className={className} {...props}>
```
tristan-orourke, line 93: Why do you single out `className`? It could simply be part of `...props`.

resources/assets/js/components/H2Components/Dialog.tsx
```diff
@@ -0,0 +1,184 @@
...
+        <div data-h2-dialog-wrapper {...props}>
+          {children}
+        </div>
+      </div>
+    </DialogContext.Provider>
+  );
+};
+
+// We expose the children components here, as properties.
+Dialog.Actions = Actions;
```
tristan-orourke, line 174: Is there any particular reason to export things like this? Wouldn't it be more conventional to just use named exports for everything?

resources/assets/js/stories/H2Components/Accordion.stories.tsx
```diff
@@ -0,0 +1,74 @@
...
+);
+
+stories.add(
+  "Accordion",
+  (): React.ReactElement => (
+    <section>
+      <Accordion
+        triggerIconPos="right"
+        data-h2-card="white, round"
+        data-h2-margin="b(bottom, 1)"
```
tristan-orourke, line 17: I like how you've set up several differently styled options. I think it would be even more useful if you made the most relevant attributes "knobs", so you can test them out from within storybook. The three accordions you have here could each be their own story, with different default knob values.
(This applies to the button and dialog stories as well.)

resources/assets/js/stories/H2Components/Button.stories.tsx
```diff
@@ -0,0 +1,29 @@
...
+);
+
+stories.add(
+  "Button",
+  (): React.ReactElement => (
+    <section>
+      <Button
+        styling="theme-1, round, medium, solid"
+        data-h2-margin="b(right, 1)"
+      >
```
tristan-orourke, line 17: I mentioned elsewhere I'm not sure we need a custom Button component. But even if we don't, its nice to have stories like as examples of how to style them!

resources/assets/js/stories/H2Components/Dialog.stories.tsx
```diff
@@ -0,0 +1,71 @@
...
+          </p>
+        </Dialog.Content>
+        <Dialog.Actions data-h2-padding="b(all, .5)" data-h2-align="b(right)">
+          <Dialog.ActionBtn
+            styling="stop, round, solid"
+            data-h2-padding="b(rl, 2) b(tb, .5)"
+            data-h2-bg-color="b(white, 1)"
+          >
+            Okay, Close
+          </Dialog.ActionBtn>
```
tristan-orourke, line 65: Can we have several stories, with examples showing two or three action buttons, and one showing how to use the ExitBtn?

tristan-orourke (changes requested):

resources/assets/js/components/H2Components/Dialog.tsx
yonikid15: `className` doesn't seem to be part of the default props type list, with the sub-components... I'm still trying find a fix for it though.

resources/assets/js/components/H2Components/Dialog.tsx
```diff
@@ -0,0 +1,220 @@
...
+  const { children } = props;
+  return (
+    <div data-h2-dialog-actions {...props}>
+      {children}
+    </div>
+  );
+};
+
+const ActionBtn: React.FunctionComponent<GeneralBtnProps> = (props) => {
+  const { id = "dialog" } = useDialogContext();
```
tristan-orourke, line 50: Is this setting "dialog" as a default value? It might be simpler to make id non-optional in DialogContext, and if there needs to be a default value, it can be defined just once in the `Dialog` component where it's passed into the context provider.

resources/assets/js/components/H2Components/utils.ts
```diff
@@ -0,0 +1,13 @@
...
+  className?: string;
+}
+
+export interface GeneralBtnProps {
+  /** The hydrogen styling for a button component */
+  buttonStyling?: string;
+  /** The type of button: "submit" | "reset" | "button" */
+  type?: "submit" | "reset" | "button";
+  /** The onClick event handler */
+  onClick?: () => void;
```
tristan-orourke, line 12: To be fully generic, the onClick definition should be `(event: React.MouseEvent<HTMLButtonElement, MouseEvent>) => void`. Callback functions that we actually pass in can ignore the event parameter.

resources/assets/js/components/H2Components/Dialog.tsx
```diff
@@ -0,0 +1,220 @@
...
+    </div>
+  );
+};
+
+const ExitBtn: React.FunctionComponent<GeneralBtnProps & GeneralProps> = (
+  props,
+) => {
+  useDialogContext(); // Ensures sub-component can only be used within the Dialog component.
+  const { buttonStyling, className, onClick, children } = props;
+  const intl = useIntl();
```
tristan-orourke, line 90: The use of react-intl makes this more specific to out Talent Cloud stack, and less useful as an H2 generic component. I still think this component should just use `{children}`, or perhaps a `hiddenText` prop, rather than include specific text.

I also don't really understand how this component is meant to be used? Like what's the difference between this and an ActionBtn, which also causes the dialog to close. It would be helpful if ExitBtn were used in one of the Storybook examples.

resources/assets/js/stories/H2Components/Accordion.stories.tsx
```diff
@@ -0,0 +1,156 @@
...
+            >
+              <p
+                data-h2-font-color={`b(${select(
+                  "Text Colors",
+                  colorOptions,
+                  "black",
+                  "Text",
+                )})`}
+              >
+                {text("Title", "Accordion TItle", "Text")}
```
mnigh, line 43: @yonikid15 change _Accordion TItle_ to _Accordion Title_

resources/assets/js/components/H2Components/Dialog.tsx
```diff
@@ -0,0 +1,220 @@
...
+  );
+};
+
+interface TriggerProps extends GeneralProps, GeneralBtnProps {
+  id: string;
+}
+
+/** This Trigger component opens the dialog and sits outside the main dialog component */
+const Trigger: React.FunctionComponent<TriggerProps> = (props) => {
+  useDialogContext(); // Ensures sub-component can only be used within the Dialog component.
```
tristan-orourke, line 149: This errors because Trigger exists outside the Dialog.

resources/assets/js/stories/H2Components/Accordion.stories.tsx
```diff
@@ -0,0 +1,151 @@
...
+                {text("Content", "Accordion Content", "Text")}
+              </p>
+            </div>
+          </Accordion.Content>
+        </Accordion>
+      </section>
+    ),
+  )
+  .add(
+    "Default",
```
tristan-orourke, line 84: What's the difference between the Accordion Card and Default stories?

tristan-orourke (changes requested):
You've addressed my previous comments (I like new Accordion storybook file) but I found a few new things. I'm still a bit confused about the ExitBtn - it might be helpful to see it in a storybook example - and I think we can't use react-intl if we want to include this in H2.

resources/assets/js/components/H2Components/Dialog.tsx
yonikid15: Ya this is meant for the close button that sits in the top right hand corner of the dialog/modal. You're right it is pretty much the same as the ActionBtn. I'll remove it then, and there will be just one button component within the Dialog.

resources/assets/js/stories/H2Components/Accordion.stories.tsx
yonikid15: The Accordion Card uses the card property giving it that box shadow effect. Maybe I should call it fancy accordion haha? Default is just a default accordion.

tristan-orourke (approved):

PR #5151, opened 2021-02-02
Task - Combine skills modal and search bar

### LINKED ISSUE #4857: Task -  Combine Explore Skills modal and Search Bar
# Description
The newer version of the Explore Skills modal combines the previous Explore Skills ui with the search bar, and includes them directly in the My Skills page.

# Images
[Updated My Skills page wireframe](https://xd.adobe.com/view/c2c5a303-1dcd-4436-bec3-eea0a3ca3523-9796/)

# Accessibility Requirements
https://www.notion.so/bitliege/Skill-Accessibility-Requirements-42a7bd9f17a04c3eb8ee333848ef454b

yonikid15:
Resolves #4857 

### Notes:
- Combines the search skills input to the find skills modal. 
- Viewable in storybook under ApplicantProfile -> FindSkillsModal.


resources/assets/js/components/SearchBar.tsx
```diff
@@ -2,21 +2,19 @@ import * as React from "react";
...
-    defaultMessage: "Search",
+    defaultMessage: "Search for skills by name:",
     description: "The label displayed for the search input field.",
   },
   searchPlaceholder: {
     id: "searchBar.search.placeholder",
-    defaultMessage: "eg. Enter a value.",
+    defaultMessage: "eg. User interface design.",
     description: "The placeholder displayed for the search input field.",
   },
```
tristan-orourke, line 19: I see you've moved this out of the ApplicantProfile directory because its _so close_ to being a totally generic search bar component. These strings are still specific to Skill search, so maybe pass them in as props?

resources/assets/js/components/FindSkillsModal.tsx
```diff
@@ -476,7 +515,7 @@ const FindSkillsModal: React.FunctionComponent<FindSkillsModalProps> = ({
               data-h2-padding="b(rl, 2) b(tb, .5)"
               data-h2-bg-color="b(white, 1)"
               onClick={() => {
-                setSkillsResults([]);
+                setNewSkills([]);
```
tristan-orourke, line 518: I just noticed that this will trigger if you click Cancel to close the dialog, but _not_ if you press Esc. Its not a big deal in this case, but its something we need to be aware of when using this version of the Dialog component.

tristan-orourke (changes requested):
The new Skills Modal looks and works great.

I don't like having the SkillSearchResults component hanging around that directly imports h2AccordionToggle instead of using the Accordion component. Since we're not using it, I think we can just delete it instead of converting it.

I have a slight concern about the "required" validation message on the search bar, but its an edge case and I'm not sure about a better alternative. If you do a search, but then erase what you've written (or clear with the x button) you see an error message. It just annoys me a little that it never goes away if you go back to using the accordion skill accordions, unless you type something out in the field again. I'm okay ignoring this, but if it bothers anyone else as well we could think about it.
image

resources/assets/js/stories/FindSkillsModal.stories.tsx
```diff
@@ -2,21 +2,21 @@ import React from "react";
...
-import { fakeSkillCategories } from "../../fakeData/fakeSkillCategories";
+import FindSkillsModal from "../components/FindSkillsModal";
+import { fakeSkill2, fakeSkill4, fakeSkills } from "../fakeData/fakeSkills";
+import { fakeSkillCategories } from "../fakeData/fakeSkillCategories";
 
-const stories = storiesOf(
-  "Applicant Profile/Find Skills Modal",
-  module,
-).addDecorator(withIntl);
+const stories = storiesOf("Components/Find Skills Modal", module).addDecorator(
```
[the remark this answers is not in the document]
yonikid15: Yea since it's going to be used on the Manager Portal as well.

resources/assets/js/components/ApplicantProfile/Skills/SkillSearchResults.tsx
```diff
@@ -154,7 +154,7 @@ export const SkillSearchResults: React.FunctionComponent<SkillSearchResultsProps
 }: SkillSearchResultsProps): React.ReactElement => {
   const intl = useIntl();
   React.useEffect((): void => {
-    h2ComponentAccordionAddTriggerEvent();
+    h2AccordionToggle();
```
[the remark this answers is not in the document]
yonikid15: I deleted it, since it's not used anywhere other then storybook.

resources/assets/js/components/FindSkillsModal.tsx
yonikid15: So, I changed it to set the new skills array to zero `setNewSkills([])` only after submission. This should trigger a re-render and if they open the add skills modal again it will set those skills they just added to "Added already". Then I removed any events if they hit cancel, so if they hit cancel and open it again they will pick up wear they left off.

yonikid15:
> The new Skills Modal looks and works great.
> 
> I don't like having the SkillSearchResults component hanging around that directly imports h2AccordionToggle instead of using the Accordion component. Since we're not using it, I think we can just delete it instead of converting it.
> 
> I have a slight concern about the "required" validation message on the search bar, but its an edge case and I'm not sure about a better alternative. If you do a search, but then erase what you've written (or clear with the x button) you see an error message. It just annoys me a little that it never goes away if you go back to using the accordion skill accordions, unless you type something out in the field again. I'm okay ignoring this, but if it bothers anyone else as well we could think about it.
> image

For the required validation I agree that the error should disappear after clicking off of it. I'm not sure of a better solution right now. We could remove the required validation, but that would mean the user can submit an empty string, which from testing returns all the results.

resources/assets/js/components/FindSkillsModal.tsx
```diff
@@ -101,6 +120,34 @@ const FindSkillsModal: React.FunctionComponent<FindSkillsModalProps> = ({
...
+        searchQuery,
+      }),
+      description: "",
+    });
+    setSkillsResults(skillMatches);
+    return Promise.resolve();
+  };
   return (
     <section>
       <Dialog.Trigger
```
tristan-orourke, line 153: What the trigger looks like seems like it might be something the parent of FindSkillsModal might want to control. The fact that there's a placeholder image here right now kind of proves that this component doesn't know what the trigger should look like. Lets replace it with a `triggerContent: React.ReactElement` prop.

resources/assets/js/stories/FindSkillsModal.stories.tsx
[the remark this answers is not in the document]
tristan-orourke: Hm, okay. I still think it makes more sense to keep it in the Applicant Profile section, than remove it from both the Applicant Profile and Manager Portal sections.

tristan-orourke (changes requested):
Looks pretty good, I just want to allow (force?) the content of the Trigger to be passed in.

resources/assets/js/components/FindSkillsModal.tsx
```diff
@@ -64,18 +69,32 @@ const messages = defineMessages({
...
   oldSkills: Skill[];
   portal: "applicant" | "manager";
+  skills: Skill[];
   skillCategories: SkillCategory[];
   handleSubmit: (values: Skill[]) => Promise<void>;
 }
 
 const FindSkillsModal: React.FunctionComponent<FindSkillsModalProps> = ({
   oldSkills,
   portal,
```
tristan-orourke, line 96: Is the portal prop really required? Seems we could replace it by simply setting `oldSkills` to an empty array.

resources/assets/js/components/FindSkillsModal.tsx
tristan-orourke: Looking closer at this, I see that the properties on the Dialog.Trigger component itself are pretty important. So to allow the parent to control what the trigger looks like, we'd either need two props, like so:
```
 <Dialog.Trigger id="findSkills" {...triggerProps}>
    {triggerContent}
 </Dialog.Trigger>
 ```
 Or allow the parent to pass in the whole thing, almost like a `{children}` slot. That would suggest treating FindSkillsModal as _just_ the Dialog component, passing in the id from the parent, and probably allowing the parent to own the Dialog.Overlay as well:
 ```
  <Dialog.Trigger id="findSkills">
    ...
 </Dialog.Trigger>
 <FindSkillsModal id="findSkills" ... />
 <Dialog.Overlay />
 ```

File: resources/assets/js/components/FindSkillsModal.tsx [hunk 7]
```diff
@@ -101,6 +120,34 @@ const FindSkillsModal: React.FunctionComponent<FindSkillsModalProps> = ({
...
+        numOfSkills: skillMatches.length,
+        searchQuery,
+      }),
+      description: "",
+    });
+    setSkillsResults(skillMatches);
+    return Promise.resolve();
+  };
   return (
     <section>
```
tristan-orourke, line 152: Sorry, something else I noticed after submitting the review - I don't think this component should include a section tag. The parent can do that if it wants.

tristan-orourke:
> For the required validation I agree that the error should disappear after clicking off of it. I'm not sure of a better solution right now. We could remove the required validation, but that would mean the user can submit an empty string, which from testing returns all the results.

I think returning all results might be okay? I think its what a user might expect, in fact. I'll check in with Jerbo. There's a chance we deliberately want to hide the full list from users, to guide them towards the highlighted ones.

tristan-orourke:
Also, is there something wrong the font of the error text here?
image

resources/assets/js/components/FindSkillsModal.tsx
yonikid15: Yup, that's good idea 👍 I think the latter of the two options is better. Have a prop called something like `triggerContent` and pass the `string | React.ReactElement` in there. However, I think the overlay should stay with the Dialog component, I don't see any advantage having the parent components having to remember to add that piece in. In fact, we could probably just get rid of the overlay sub-component completely and just add it to the main Dialog component, so we don't have to worry about it.

tristan-orourke: That's a good point! I'm all in favour of adding the overlay to the main Dialog component, as long as we can confirm those no issue with having multiple Overlays in a page.

yonikid15:
> Also, is there something wrong the font of the error text here?
> image

@tristan-orourke @bitliege  It looks like hydrogen's `span` class selector is interfering with clone's. The span's font-family selector in hydrogen is set to inherit. However, I don't think there is any parent element to inherit from causing it to set it to the default Times New Roman. A quick solution would be to change the span's font-family in hydrogen from `inherit` to `$font-default`. Or, we can create new Input element using hydrogen.

resources/assets/js/components/FindSkillsModal.tsx
yonikid15: Hmm, but when a manager tries to reset the skill filters they previously set, we want to be able to know the previous skills they picked with the `oldSkills` prop. I added the portal prop to conditionally add that added filters section, in the manager's find skills modal, and also to remove the disabled "Already Added" button.

tristan-orourke:
> @tristan-orourke @bitliege It looks like hydrogen's `span` class selector is interfering with clone's. The span's font-family selector in hydrogen is set to inherit. However, I don't think there is any parent element to inherit from causing it to set it to the default Times New Roman. A quick solution would be to change the span's font-family in hydrogen from `inherit` to `$font-default`. Or, we can create new Input element using hydrogen.

Ah, yeah, we should probably build an H2 input component.

resources/assets/js/components/FindSkillsModal.tsx
tristan-orourke: Okay, so there will be other things later that only appear in the manager version? Makes sense.

tristan-orourke:
> > For the required validation I agree that the error should disappear after clicking off of it. I'm not sure of a better solution right now. We could remove the required validation, but that would mean the user can submit an empty string, which from testing returns all the results.
> 
> I think returning all results might be okay? I think its what a user might expect, in fact. I'll check in with Jerbo. There's a chance we deliberately want to hide the full list from users, to guide them towards the highlighted ones.

I talked to Jerbo, we agreed that a good compromise would be to remove the `required` validation error, but a special case to the search handler function so that searching with an empty string just clears the results. I have the code ready, I'll push it up in a minute.

tristan-orourke (approved):

PR #5188, opened 2021-02-05
Bugfix - Removing Applicant Skills deletes ExperienceSkills

### LINKED ISSUE #5159: Bug - When deleting Applicant Skills, ensure associated ExperienceSkills are deleted
# Description
In ApplicantSkillsController, when we update an applicants skills, we `sync` the request data with the list of Applicant Skills. Some Applicant Skills may be deleted here. 
According to [this wireframe](https://xd.adobe.com/view/c2c5a303-1dcd-4436-bec3-eea0a3ca3523-9796/) removing an Applicant Skill will also remove that skill from any Experiences on your profile. This means that we need to ensure any Skills removed during the `sync` also cause associated ExperienceSkills to be (soft) deleted.

tristan-orourke:
Resolves #5159 .

### Notes:
- Added a method to Applicant eloquent model which returns a query (not the results) which gets all ExperienceSkills belonging to that applicant.

### DIFF
--- a/app/Models/Applicant.php
+++ b/app/Models/Applicant.php
@@ -40,6 +40,8 @@ use App\Traits\TalentCloudCrudTrait as CrudTrait;
  * @property \Illuminate\Database\Eloquent\Collection $experiences_education
  * @property \Illuminate\Database\Eloquent\Collection $experiences_award
  * @property \Illuminate\Database\Eloquent\Collection $experiences_community
+ *
+ * @method \Illuminate\Database\Query\Builder experienceSkillsQuery
  */
 class Applicant extends BaseModel
 {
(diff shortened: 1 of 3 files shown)


app/Models/Applicant.php
```diff
@@ -170,4 +172,69 @@ public function experiences_community() //phpcs:ignore
...
+
+            $query->orWhere(function ($query) use ($applicantId) {
+                $query->whereIn('id', function ($query) use ($applicantId) {
+                    $query->select('experience_skills.id')->from('experience_skills')->join('experiences_work', function ($join) use ($applicantId) {
+                        $join->on('experience_skills.experience_id', '=', 'experiences_work.id')
+                            ->where('experience_skills.experience_type', 'experience_work')
+                            ->where('experiences_work.experienceable_type', 'applicant')
+                            ->where('experiences_work.experienceable_id', $applicantId);
+                    });
+                });
```
yonikid15, line 200: These all look pretty similar, other then the experience types. Maybe we should abstract this out into a function. Although, I'm trying now and it doesn't seem to be cooperating.  😅

yonikid15: This is what I have so far.. However, the tests assertSoftDeleted test fails.
```php 

public function experienceSkillsQuery()
    {
        $applicantId = $this->id;
        $getExperienceSkills = function ($query, $experiences, $experienceType, $applicantId): void {
            $query->whereIn('id', function ($query) use ($experiences, $experienceType, $applicantId): void {
                $query->select('experience_skills.id')
                    ->from('experience_skills')
                    ->join($experiences, function ($join) use ($experiences, $experienceType, $applicantId): void {
                        $join->on('experience_skills.experience_id', '=', "{$experiences}.id")
                        ->where('experience_skills.experience_type', $experienceType)
                        ->where("{$experiences}.experienceable_type", 'applicant')
                        ->where("{$experiences}.experienceable_id", $applicantId);
                    });
            });
        };

        return ExperienceSkill::where(function ($query) use ($getExperienceSkills, $applicantId): void {
            // A single where clause wraps 5 OR WHERE clauses.
            // Each WHERE clause gets all the ExperienceSkills linked to a particular Experience type (and this Applicant).

            $query
                ->orWhere($getExperienceSkills($query, 'experiences_work', 'experience_work', $applicantId))
                ->orWhere($getExperienceSkills($query, 'experiences_personal', 'experience_personal', $applicantId))
                ->orWhere($getExperienceSkills($query, 'experiences_education', 'experience_education', $applicantId))
                ->orWhere($getExperienceSkills($query, 'experiences_award', 'experience_award', $applicantId))
                ->orWhere($getExperienceSkills($query, 'experiences_community', 'experiences_community', $applicantId));
        });
    }

```

yonikid15 (approved):
Nice work 👍  I tried moving the `orWhere` clause function, into it's own function, so it can reused for each `experience_type`. However, the test is failing. I'll approve this, but take a look and let me know what you think.

app/Models/Applicant.php
tristan-orourke: Great suggestion! The trick was to write a function that returns a function. If you like my solution, merge it in!

yonikid15 (approved):
Looks good 👍

PR #5189, opened 2021-02-05
Task - show password button accessibility

### LINKED ISSUE #5160: Add labels to show/hide password buttons on login and registration forms
Or hidden text that will be picked up by screen readers.

"Show/hide" should be sufficient.

yonikid15:
Resolves #5160

### Notes:
- I wasn't able to add any text into the button without breaking the styling of the show/hide icons. We could fix this by changing [Clone's password component](https://designwithclone.ca/components/input/password/). However, I think I found another solution using a combo of the `aria-label` and `aria-pressed` attributes. When the user clicks the button, the screen reader will hopefully read, "Show password button pressed".

### DIFF
--- a/resources/views/auth/register.html.twig
+++ b/resources/views/auth/register.html.twig
@@ -106,9 +106,9 @@
 										<span>{{ trans('forms.required') }}</span>
 										<div>
 											<input id="password" required type="password" name="password" autocomplete="off"/>
-											<button type="button">
-												<i class="fas fa-eye"></i>
-												<i class="fas fa-eye-slash"></i>
+											<button class="showPassword" aria-label={{register.show_password}} aria-pressed="false" type="button">
+												<i class="fas fa-eye" aria-hidden="true"></i>
+												<i class="fas fa-eye-slash" aria-hidden="true"></i>
 											</button>
 										</div>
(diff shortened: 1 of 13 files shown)


resources/views/auth/register.html.twig
```diff
@@ -106,9 +106,9 @@
 										<span>{{ trans('forms.required') }}</span>
 										<div>
 											<input id="password" required type="password" name="password" autocomplete="off"/>
-											<button type="button">
-												<i class="fas fa-eye"></i>
-												<i class="fas fa-eye-slash"></i>
+											<button class="showPassword" aria-label={{register.show_password}} aria-pressed="false" type="button">
+												<i class="fas fa-eye" aria-hidden="true"></i>
+												<i class="fas fa-eye-slash" aria-hidden="true"></i>
```
tristan-orourke, line 111: You've set aria-hidden only on these icons. Should we set aria-hidden on the other show password buttons as well?

tristan-orourke (changes requested):

resources/views/auth/register.html.twig
yonikid15: Yes, missed that thanks

tristan-orourke (approved):

PR #5207, opened 2021-02-10
Bugfix - test get applicant skills unit test

yonikid15:
PHP Unit test occasionally fails with the following:

image

### DIFF
--- a/tests/Feature/Api/ApplicantSkillsControllerTest.php
+++ b/tests/Feature/Api/ApplicantSkillsControllerTest.php
@@ -24,7 +24,7 @@ class ApplicantSkillsControllerTest extends TestCase
         $response = $this->actingAs($applicant->user)->json('get', route('api.v1.applicant.skills.index', $applicant));
         $response->assertOk();
         $response->assertJson([
-            'skill_ids' => $skills->sort()->pluck('id')->all()
+            'skill_ids' => $skills->sortBy('id')->pluck('id')->all()
         ]);
     }
 
--- a/tests/Feature/Api/ApplicantSkillsControllerTest.php
+++ b/tests/Feature/Api/ApplicantSkillsControllerTest.php
@@ -6,7 +6,6 @@ use App\Models\Applicant;
 use App\Models\Skill;
 use Illuminate\Foundation\Testing\RefreshDatabase;
 use Tests\TestCase;
-use Illuminate\Support\Facades\Log;
 
 class ApplicantSkillsControllerTest extends TestCase
 {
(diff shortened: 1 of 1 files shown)


tests/Feature/Api/ApplicantSkillsControllerTest.php
```diff
@@ -15,7 +15,7 @@ public function testGetApplicantSkills(): void
     {
         $applicant = factory(Applicant::class)->create();
         $skills = Skill::inRandomOrder()->limit(5)->get();
-        $applicant->skills()->sync($skills->pluck('id'));
+        $applicant->skills()->sync($skills->sortBy('id')->pluck('id'));
```
mnigh, line 18: @yonikid15 could it be failing because there is no id to sort by in the array? what about `sort()` instead?

yonikid15: There's no id in the collection of skill models?

mnigh:
Code coverage test fails arbitrary coverage drop allowed for success value, 2.5%, by 0.17% (2.67%).

tests/Feature/Api/ApplicantSkillsControllerTest.php
mnigh: No, because it's no longer a collection after `pluck('id')`, only an array of the values. 

> [The pluck method retrieves all of the values for a given key.](https://laravel.com/docs/8.x/collections#method-pluck)

tests/Feature/Api/ApplicantSkillsControllerTest.php
```diff
@@ -24,7 +24,7 @@ public function testGetApplicantSkills(): void
         $response = $this->actingAs($applicant->user)->json('get', route('api.v1.applicant.skills.index', $applicant));
         $response->assertOk();
         $response->assertJson([
-            'skill_ids' => $skills->sortBy('id')->pluck('id')->all()
+            'skill_ids' => $skills->sort()->pluck('id')->all()
```
tristan-orourke, line 27: Any reason it was changed _to_ sortBy on line 19 and _away from_ sortBy on line 27?

tristan-orourke (changes requested):

tests/Feature/Api/ApplicantSkillsControllerTest.php
mnigh: My thinking was that the collection in $skills on line 18 contains an array with items that each have an `id` key, whereas on line 27, $skills is only an array of the values of the `id` key and therefore there would be no `id` key to sort by, but I believe I am wrong since it's only an array of the values of the `id` key _after_ the `sortBy` method is called, since the `pluck` method is called after. My fault.

tristan-orourke (approved):

PR #5211, opened 2021-02-11
Resource Hook api errors will now trigger Error Toasts

### LINKED ISSUE #5214: Task - Set up a way for resource hook API request errors to trigger Error Toast
# Description
API requests that go through the Redux asyncActions are set up so they trigger an Error Toast if an error happens. Ensure that requests going through resourceHooks trigger a similar error toast. 

Probably want to use a React Context object to pass errors to the toast component, instead of Redux.

tristan-orourke:
Resolves #5214.

Creates a new version of the ErrorToast component, using the H2 Alert, and using a a queue of errors accessed from a React Context instead of Redux.

### Notes:
- Wrote an easy to use handleError function in `apiResourceHooks` which constructs an error message and pushes it to the Error Context. If no ErrorContext.Provider exists in the component hierarchy, nothing will happen (but it won't throw an exception or anything).
- Previous toast component renamed to ErrorToastRedux.
- Unlike Redux error store, this Error Context stores a queue of errors. Dismissing one error causes the next one to appear, if any.
- I wrote some quite clever code (I think ;) to dismiss the error after a timer but also allow manual dismissal. But I removed it in a later commit after confirming that the automatic dismissal isn't great from an accessibility standpoint.
- It might be useful, accessibility wise, for the toast to grab focus. I'm not sure how to do that.
- I'm not sure how to write an automated test for this. I tested it locally by adding some middleware to the _applicant.skills.index_ route that forced it to fail. I set line 961 of _routes/web.php_ to:
```->middleware('can:view,applicant')->middleware('can:create,App\Models\User')```

### CODE REVIEW

File: resources/assets/js/components/ErrorToast.tsx [hunk 1]
```diff
@@ -1,94 +1,57 @@
...
-  useEffect((): (() => void) => {
-    const timer = setInterval((): void => {
-      dispatchClearErrors();
-      clearInterval(timer);
-    }, 3500);
-    return (): void => {
-      clearInterval(timer);
-    };
-  }, [dispatchClearErrors, error]);
+const TOAST_SELF_DISMISS_TIMER = 3500;
```
yonikid15, line 6: `TOAST_SELF_DISMISS_TIMER` isn't being used anywhere. Same with `useEffect` and `useRef` in the imports section.

yonikid15 (approved):
Nice work 👍 I just fixed a couple of the comments I made. Also, the alert component styling was off because it was missing the `data-h2-alert-wrapper="toast` property the wraps the alert component, so I added it in.

PR #5268, opened 2021-02-19
Bugfix - Save Experience start_date, end_date, awarded_date correctly

### LINKED ISSUE #5158: Bug - Dates submitted in Experiences change every time the Experience is updated
# Description
When the user first selects a date on an Experience form (eg start date, end date, date awarded) it is saved correctly in the database. However, because of a discrepancy between the UTC timezone all dates are saved with on the server, and the user's browser's local timezone, the date may appear differently on the client side. And if the user edits that experience, the new - incorrect - date will be saved to the backend.

# Steps to Reproduce
1. Go to the Experience page of the profile or the Experience step of the application.
2 Click "Add Work Experience"
3. Fill out the Experience form, setting the Start Date as April 20.
4. Click Save Experience
5. Find the new Experience in the accordions. Note that the Start Date is April 19.
6. Click Edit Experience on that accordion.
7. Without changing anything, click Save Experience again
8. Not that the Start Date is now April 18.

# Notes
The date shift is happening because of the way we convert between Date and Datetime objects in the background between the client and server, and because our Experience models, unlike most of our other models, do not save _time_ portion of Datetime objects in the database.

1. User picks a date in a Date input element (eg April 20 2020)
2. Client-side js represents this as a Datetime, setting the time to midnight _local time_.
3. When client sends update request to server, Datetime data includes embedded timezone (eg `2020-04-20T00:00:00-04:00`)
4. Server interprets the date correctly, but when it is saved in database, the time and timezone data is stripped. (eg saved as `2020-04-20`)
**So far, so good. But...**
5. Server responds to the api request with an updated version of the Experience model. When Laravel retrieves the date from the database, it represents it as a Datetime again. The time defaults to midnight _Universal Standard Time (UTC)_. (eg `2020-04-20T00:00:00+00:00`)
6. Client recieves the UTC date, and converts it to the browser's local time (eg `2020-04-19T20:00:00-04:00`)
7. The date is now displayed to the user as April 19 2020.
8. If the user edits this experience, the date appears in the date picker element as April 19 2020. If they save the form, return to step 2.

# Solutions & Concerns
- The problem might just go away if we make sure to save all dates as timestamps and datetimes in the database and the eloquent models respectively.
- Is there a better way of representing dates where we don't care about the time, and keeping that synced between front- and back-end?
- This issue could exist elsewhere, if any other models are saving dates without time data in the database.

mnigh:
Resolves #5158.


resources/assets/js/components/Application/ExperienceModals/PersonalExperienceModal.tsx
```diff
@@ -280,7 +276,7 @@ export const newPersonalExperience = (
   description: "",
   is_shareable: false,
   is_active: false,
-  start_date: new Date(),
+  start_date: dayjs().format("YYYY-MM-DD"),
```
tristan-orourke, line 279: It might be nice to abstract this away to a `newDateString()` helper function. I know its super simple, but I like not having to remember or repeat the format string in multiple places.

tristan-orourke (approved):
Despite my one suggestion of making a DateString helper function, everything looks good! Merge whenever you want.

app/Http/Resources/Experience.php
```diff
@@ -33,6 +34,20 @@ public function toArray($request)
             [
                 'type' => $type,
                 'experience_skills' => JsonResource::collection($this->whenLoaded('experience_skills')),
+            ],
+            [
+                'start_date' => $this->when($this->start_date, function () {
+                    $start_date = new Date($this->start_date);
```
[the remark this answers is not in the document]
mnigh: It is a date object or null, but I've refactored this to make sure that only the necessary key/value pairs are returned for each experience type.

app/Http/Resources/Experience.php
```diff
@@ -17,16 +17,45 @@ public function toArray($request)
...
-            $translations = [
-                'award_recipient_type' => $this->award_recipient_type->getTranslations('name'),
-                'award_recognition_type' => $this->award_recognition_type->getTranslations('name'),
-            ];
+        $dates = [];
+        switch ($type) {
+            case 'experience_work':
+                $dates = [
+                    'start_date' => $this->when($this->start_date, $this->start_date->format('Y-m-d')),
+                    'end_date' => $this->when($this->end_date, $this->end_date->format('Y-m-d')),
```
tristan-orourke, line 25: Oops! Turns out this causes an error when end_date is null. I guess `->when()` doesn't protect against null values? (To be fair, it seems to be intended for something different, ie relationships which may be loaded or not.)

tristan-orourke (changes requested):

tristan-orourke (approved):

PR #5283, opened 2021-02-22
Update user password length rules

### LINKED ISSUE #5267: Increase minimum password length from 8 to 9 characters
As recommended by the security consultant.

mnigh:
Resolves #5267.

### Notes:
- Changes minimum rule for password from 8 characters to 9 characters
- Fixes missing minimum rule on password update
- Adds maximum rule for password of 100 characters


app/Services/Validation/PasswordValidator.php
```diff
@@ -15,7 +15,7 @@ public function rules()
     {
         return [
             'password' => [
-                'min:8',
+                'min:9',
                 'regex:/^.*(?=.{3,})(?=.*[a-z])(?=.*[A-Z])(?=.*[0-9])(?=.*[!@#$%^&*]).*$/',
```
yonikid15, line 19: We should replace the `regex` rule here with `PasswordFormatRule` to keep it consistent with the entire site. Also, I don't think this validator is being referenced anywhere, now that I'm looking. If that's the case you can just delete the file.

app/Http/Controllers/Auth/ResetPasswordController.php
```diff
@@ -83,7 +83,7 @@ protected function rules()
             'email' => 'required|email',
             'password' => [
                 'required',
-                'min:8',
+                'min:9',
```
yonikid15, line 86: If you go to `PasswordFormatRule.php` and look for the [regex](https://github.com/GCTC-NTGC/TalentCloud/blob/95a834e26598b42729d2147a2379717aac2c2ed4/app/Services/Validation/Rules/PasswordFormatRule.php#L14) under `passwordPattern`. It says that it accepts a minimum of 8 characters. I'm not sure if the `min` validation rule here takes priority over it, but maybe we should change the regex just incase it doesn't.

yonikid15 (changes requested):
Just a few things left. Also the [password validation text](https://github.com/GCTC-NTGC/TalentCloud/blob/95a834e26598b42729d2147a2379717aac2c2ed4/resources/lang/en/common/auth/register.php#L29) should get updated as well to inform the user they need to have a min of 9 characters.

tristan-orourke:
As we discussed recently on slack, lets add a max password length while we're at this, just to avoid anything too absurd. Max length 50? 100?

app/Http/Controllers/Auth/ResetPasswordController.php
mnigh: @yonikid15 I'm not sure one takes priority over the other, I think both rules are applied, based on testing.

yonikid15 (approved):
Nice Work 👍

PR #5284, opened 2021-02-22
Collapse redux logs that appear in console

tristan-orourke:
This is a QOL change for devs, as the logs don't appear in production anyway.


mnigh (approved):
Nice and clean, thanks @tristan-orourke!

PR #5318, opened 2021-02-26
Temporarily remove description from client-side SkillCategory

tristan-orourke:
Code is commented out instead of deleted becuase we want to add description to backend model and then restore this.


yonikid15 (approved):
Looks good

PR #5320, opened 2021-02-26
Log non-GET Admin requests

### LINKED ISSUE #5215: Task - Audit Logs - Log whenever an Admin makes a POST, PUT or DELETE request.
# Description
For audit purposes, we should log whenever an admin makes a change on the server, because admins have the ability to modify data belonging to other users.
We can add middleware that runs on every request. If the current user is logged in as an admin, and its not a GET request, log the request url, verb, and body.
This can replace the logs we currently have in CheckRole.php and BasePolicy.php.

mnigh:
Resolves #5215.

### DIFF
--- a/app/Http/Middleware/LogAction.php
+++ b/app/Http/Middleware/LogAction.php
@@ -1,0 +1,39 @@
+<?php
+
+namespace App\Http\Middleware;
+
+use Closure;
+use Illuminate\Support\Facades\Auth;
+use Illuminate\Support\Facades\Log;
+
+class LogAction
+{
+
+    /**
+     * Handle an incoming request.
+     *
+     * @param  \Illuminate\Http\Request  $request
+     * @param  \Closure  $next
+     * @return mixed
+     */
+    public function handle($request, Closure $next)
+    {
+        if (Auth::check() && Auth::user()->isAdmin() && $request->method() !== 'GET') {
+            parse_str($request->getContent(), $body);
+            $bodyKeys = [];
+            foreach ($body as $key => $value) {
+                array_push($bodyKeys, $key); // Only include keys from request body to avoid logging any sensitive values.
+            }
+            $logArray = [
+                'message' => 'Admin non-GET request',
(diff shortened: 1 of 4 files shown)


app/Http/Middleware/LogAction.php
```diff
@@ -0,0 +1,39 @@
...
+     * @return mixed
+     */
+    public function handle($request, Closure $next)
+    {
+        if (Auth::check() && Auth::user()->isAdmin() && $request->method() !== 'GET') {
+            parse_str($request->getContent(), $body);
+            $bodyKeys = [];
+            foreach ($body as $key => $value) {
+                array_push($bodyKeys, $key); // Only include keys from request body to avoid logging any sensitive values.
+            }
```
tristan-orourke, line 26: $request->getContent() seems to return just a single string, which gets logged as one giant key. Instead of getContent, try using $request->input(), which matches how we access request data in our controllers.

app/Http/Middleware/LogAction.php
```diff
@@ -0,0 +1,39 @@
...
+            $logArray = [
+                'message' => 'Admin non-GET request',
+                'admin_user_id' => $request->user()->id,
+                'url' => $request->url(),
+                'verb' => $request->method(),
+                'body' => $bodyKeys
```
tristan-orourke, line 32: Maybe change this key from "body" to "bodyKeys", to be more clear about whats actually being logged.

tristan-orourke (changes requested):
Logging the keys but not values seems like a good compromise 👍 
Its not quite working like that at the moment - just need to access the body differently.

app/Http/Middleware/LogAction.php
mnigh: @tristan-orourke I've changed it to `body_keys` to match the snake case of `admin_user_id`. If you'd prefer camel case, let me know and I can the keys accordingly, thanks.

tristan-orourke (approved):

PR #5366, opened 2021-03-02
Replace Travis CI badge with GitHub Actions badge

mnigh:
### Notes:
- Travis CI badge no longer relevant since the service is no longer used.


gggrant (approved):
Stellar work. RIP Travis.

PR #5376, opened 2021-03-03
Log data associated with an application on status change

### LINKED ISSUE #5216: Task - Audit Logs - Log all data associated with an application when the status changes
# Description
We currently log when an application status changes (ie when it is submitted) but we log simply the id. We should now include all of the data associated with the application, including experiences and experienceSkills.

mnigh:
Resolves #5216.

- Additionally, adds the PHP extension `pgsql` to tests.yml to account for [recent change](https://github.com/shivammathur/setup-php/issues/425) to `ubuntu-latest`.

### DIFF
--- a/app/Listeners/ApplicationStatusChanged.php
+++ b/app/Listeners/ApplicationStatusChanged.php
@@ -11,6 +11,7 @@ use Illuminate\Support\Facades\Auth;
 use Illuminate\Database\Eloquent\Builder;
 use App\Models\Lookup\CriteriaType;
 use App\Models\Lookup\SkillType;
+use App\Http\Resources\Experience as ExperienceResource;
 
 class ApplicationStatusChanged
 {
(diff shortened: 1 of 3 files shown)


app/Listeners/ApplicationStatusChanged.php
```diff
@@ -38,17 +42,47 @@ public function handle(ApplicationSaved $event)
...
+
             $applicationText = '{id=' . $freshApplication->id . '}';
+            $jobPosterText = '{job_poster_id=' . $freshApplication->job_poster_id . '}';
             $statusText = '{' . $freshApplication->application_status->name . '}';
-
-            Log::notice('Application status changed: application ' . $applicationText . ' has been changed to ' . $statusText . ' by user ' . $userText);
+            $logArray = [
+                'message' => 'Application status changed: application ' . $applicationText . ' for ' . $jobPosterText . ' has been changed to ' . $statusText . ' by user ' . $userText,
+                'application' => $freshApplication->toArray(),
+                'experiences' => $experiences
```
tristan-orourke: Lets rename this to experience_skills, and include Experiences as well.

tristan-orourke (changes requested):
This all looks good, but I think we should include Experiences along with ExperienceSkills, for completeness.
The experienceSkillsQuery function in app/Models/Applicant.php builds a query that retrieves all experiences (of all 5 types) for an applicant. A small bit of modification would be required to have it retrieve experiences for an Application as well (changing experienceable_type from "applicant" to "application").

tristan-orourke (approved):

PR #5384, opened 2021-03-04
Add goc, alert banners for application congrats view

### LINKED ISSUE #5342: Bug - Application Timeline - Missing banners in Thanks for Applying Header
# Description
Job application is missing Government of Canada banner and Alert banner.

# Steps to Reproduce
1. Apply for a job poster
2. Click through steps
3. Submit application
4. Observe missing banners

# Images
Missing banners

mnigh:
Resolves #5342.

### DIFF
--- a/resources/views/layouts/base-test.html.twig
+++ b/resources/views/layouts/base-test.html.twig
@@ -13,6 +13,8 @@
         {% include "common/head" %}
     {# /head #}
     {# body #}
+        {% include "common/goc" %}
+        {% include "common/alert" %}
         {# Page Header #}
         {% block header %}{% endblock %}
         {# Skip to Content Link #}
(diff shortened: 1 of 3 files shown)


resources/views/layouts/base-test.html.twig
```diff
@@ -13,6 +13,8 @@
         {% include "common/head" %}
     {# /head #}
     {# body #}
+        {% include "common/goc" %}
+        {% include "common/alert" %}
```
yonikid15, line 17: Instead of using the `base-test.html.twig` template on the congrats page (which was meant for testing), we should be using the main `base.html.twig` template. Then you'll also need to add the header block (`{% block header %}{% endblock %}`) [above the body block](https://github.com/GCTC-NTGC/TalentCloud/blob/d15f5946cd2feb05f0b95619f81a9bbd8d122599/resources/views/layouts/base.html.twig#L19) in the `base.html.twig` template.

yonikid15 (changes requested):

yonikid15 (approved):

PR #5386, opened 2021-03-04
Add message for no skills in sort by group

### LINKED ISSUE #5337: Bug - Applicant Profile - Message for empty skills groups
# Description
When there are no skill categories added, there is no indication of why nothing appears. Either a message stating that there are no groups available should be displayed or the Group tab should be hidden.
image

# Steps to Reproduce
Steps to reproduce the behaviour.
1. Go to Applicant Profile
2. Click on My Skills
3. Scroll down to Your Skills
4. See **Sort By:** section

mnigh:
Resolves #5337.


yonikid15 (approved):
Approving this for now, however we need to refactor the way we display the loading state messages. I'm seeing a flash of the empty list message before the skills finish loading. Matt is making an issue.

PR #5417, opened 2021-03-11
Improve modal accessibility on Application and Profile

### LINKED ISSUE #5161: Applicant Timeline - Modal accessibility
- Use H1 elements for the modal title.
- When the modal opens, focus should be on the H1.
- Fix keyboard trap (cannot cycle through modal contents with tab).
- Ensure all form elements are focusable.
- Check the date picker, possibly causing some of these issues.

tristan-orourke:
Resolves #5161.

The main changes were:
- ensuring all inputs in the modal forms are captured in the tab loop
- Ensure the modal title receives focus when a modal opens
- Ensure the button that opened the modal receives focus when a modal closes. NOTE: If a modal opened without a button being pressed, the first focusable element in the document body receives focus (typically the skipLink). 

This applied to Experience modals on the Experience page of the profile and Experience step on Application, and to Skill Definition modals on the Skill step of the application.

### Notes:
- We talked about replacing these modals (which use Clone) with new components built with H2, but upgrading the old modals turned out to be easier.
- I added a `getFocusableElements` helper function to `helpers/forms.ts` which returns all focusable elements in the document body, or descendent of a particular element (eg a modal element).


resources/assets/js/components/Application/Skills/SkillsPage.tsx
```diff
@@ -80,6 +81,17 @@ export const SkillsPage: React.FunctionComponent<SkillsPageProps> = ({
...
+    experienceSkillsToUpdate: ExperienceSkill[],
+  ): Promise<void> => {
+    const result = await dispatch(
+      batchUpdateExperienceSkills(experienceSkillsToUpdate),
+    );
+    if (!result.error) {
+      return Promise.resolve();
+    }
+    return Promise.reject(result.error);
+  };
```
yonikid15, line 94: Thanks for handling this 👍 @mnigh made an [issue](https://github.com/GCTC-NTGC/TalentCloud/issues/5422) for it yesterday, after we noticed it.

yonikid15 (changes requested):
Nice work, just a few more problems to resolve:
- The Experience modals focusable elements are in the within the tab loop even when they're closed. 
- On the profile skills page, when I tried deleting a skill it works fine. However, when deleting another skill right after doesn't trigger the confirmation modal to popup but instead just deletes it. (Just realized this bug hasn't been discovered yet 😅)
- I'm not sure if this is the best way to do this, but after deleting an element in a list (experience or skill) shouldn't we focus on the next/previous element in the list. Right now it focuses on the "skip to main content" button. However, I'm not sure what best option is.

tristan-orourke:
I've fixed the issue of closed modal elements receiving tab focus by setting a "visibility: hidden" style. 

I have discovered a similar issue where it's possible to _leave_ the modal without closing it when using a screen reader, by using any keyboard navigation besides the tab key (eg arrow keys, hotkeys for buttons or headers, etc). 
As far as I can tell, the only solid way to address this is to put all the modals in single root div, and set `aria-hidden: true` on every other root div when a modal opens. This is involved enough to deserve its own issue, I think.

resources/assets/js/components/Modal.tsx
```diff
@@ -110,6 +140,7 @@ export default function Modal({
         role="dialog"
         ref={modalRef}
         className={className}
+        style={{ visibility: visible ? "visible" : "hidden" }}
```
yonikid15: There's a clone property to set an element's visibility to hidden: `data-c-visibility="hidden"`. I can make a quick change since I have this open.

yonikid15 (approved):
Looks good 👍

PR #5420, opened 2021-03-11
Update Application Timeline Skills Sidebar order

### LINKED ISSUE #5351: Bug - Application Timeline: Skills - Incorrect order of skills on sidebar
# Description
Occasionally, the order of the skills displayed in the sidebar will not match the order of the skills displayed on the page.
application-timeline-skill-step-sidebar-incorrect-order

# Steps to Reproduce
1. Create an application.
2. Go to the skills step.
3. See error.

mnigh:
Resolves #5351.


tristan-orourke (changes requested):
That seems to have ensured the sidebar and main skills appear in the same order, but can we order both alphabetically?

tristan-orourke (approved):
I made one small change, to use the localizeField utility for skill name comparison, but otherwise its all good.

PR #5431, opened 2021-03-12
When snapshotting profile for Application, save only content relevant to Application

### LINKED ISSUE #5389: Bug - We should only save relevant experience from profile to Application
# Description
`JobApplication->saveProfileShanpshotTimeline()` creates immutable copies of applicant experiences (and experience skills) and saves them to the application object. Ideally, it would only copy experiences and experienceSkills that are relevant to the actual job. It would make questions like analyzing the data later easier.

tristan-orourke:
Resolves #5389.


mnigh (approved):
@tristan-orourke Nice work!

PR #5450, opened 2021-03-16
Log email and timestamp for failed login

### LINKED ISSUE #5251: Task - Audit Logs - Log failed sign-in attempts
Log failed sign-in attempts so we can investigate if something is going wrong with sign-in.

If possible, log the reason it failed, but otherwise probably just email and timestamp.

Should probably keep this light, if we're getting spammed, don't wanna overload the logs.

mnigh:
Resolves #5251.

### DIFF
--- a/app/Listeners/LogFailedLogin.php
+++ b/app/Listeners/LogFailedLogin.php
@@ -4,7 +4,6 @@ namespace App\Listeners;
 
 use Illuminate\Auth\Events\Failed;
 use Illuminate\Support\Facades\Log;
-use Jenssegers\Date\Date;
 
 class LogFailedLogin
 {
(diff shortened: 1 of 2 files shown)


app/Listeners/LogFailedLogin.php
```diff
@@ -0,0 +1,30 @@
...
+
+    /**
+     * Handle the Failed event.
+     *
+     * @param  Failed  $event
+     * @return void
+     */
+    public function handle(Failed $event) : void
+    {
+        Log::notice('Login failed by email ' . $event->credentials['email'] . ' at ' . Date::now()->format('c'));
```
tristan-orourke: Every line in the log automatically includes date and time, so no need to add it manually.

tristan-orourke (changes requested):
It's just redundant to log the date manually.

tristan-orourke (approved):

PR #5475, opened 2021-03-18
Don't show unrelated skills on ApplicationPreview

### LINKED ISSUE #5402: Bug - Experience accordions on Application Preview shouldn't mention other skills on profile
# Description
Experience accordions in the Application list the skills related to the job, but also display the number of other unrelated skills the user has connected to this experience. This helps the user understand the distinction between their profile and their application, but is unhelpful when a manager is viewing their application later.

The reference to other skills on profile should be removed in the ApplicationPreview, which is how managers view applications, and how applicants view their complete applications.

If there are no relevant skills, include a null state message.

tristan-orourke:
Resolves #5402.

### Notes:
- @mnigh was also working on this ticket, my mistake 😬


yonikid15 (approved):
Looks good

PR #5486, opened 2021-03-22
Hide skills list until applicant skills have loaded at least once

### LINKED ISSUE #5392: Bug - Applicant Profile - Skills list indicates no skills during loading state
# Description
While the My Skills page of the Applicant Profile is loading, the list of skills section shows an empty skills message even if the applicant has skills attached to their profile. There should probably be a loading message or some type of loader on the list to indicate that the list is not in its loaded state.

Hide Skill List component if skills are loading AND skills list is empty.

# Steps to Reproduce
1. Go to My Profile
2. Click on My Skills
3. Scroll down to list of skills

# Images
Screen Shot 2021-03-05 at 15 19 21

tristan-orourke:
Resolves #5392.

### Notes:
- Added a `initialRefreshFinished` property to the return value of the web resource hooks. This becomes true when the initial request completes (or starts as true if the initial request is skipped). This should be useful in many similar situations, where we want to show a loading state on first load but not subsequent ones.


resources/assets/js/components/ApplicantProfile/Skills/ProfileSkillsPage.tsx
```diff
@@ -52,6 +52,12 @@ export const ProfileExperiencePage: React.FC<{ applicantId: number }> = ({
     });
   };
 
+  // The skills list should be only appear after skills have have loaded at least once.
```
mnigh: Two _haves_ don't make a whole. 😉

resources/assets/js/hooks/webResourceHooks/singleResourceHook.ts
```diff
@@ -108,6 +116,7 @@ export type UseResourceReturnType<T> = {
   value: T;
   status: ResourceStatus;
   error: undefined | Error | FetchError;
+  initialRefreshFinished: boolean; // Becomes true after the intial request is fulfilled or rejected. NOTE: if initial fetch is skipped, this will be set to true immediately.
```
mnigh: Change intial to initial.

mnigh (changes requested):
@tristan-orourke A few trivial changes left: nice work!

yonikid15 (approved):
lgtm 👍

mnigh (approved):
👍

Issue #5463, opened 2021-03-17
Task - Refactor H2 components to work without H2 scripts


tristan-orourke:
# Description
These components will be easier to use in the future if we refactor them to be more self-contained and use React conventions, instead of relying on outside scripts.

- Dialog
- Accordion
- Alert
- Form
- Select input

yonikid15:
#5497 already handles the Dialog and Accordion components.

PR #5497, opened 2021-03-23
Task - Improve find skills dialog accessibility

### LINKED ISSUE #5249: Ensure Find Skills modal meets all accessibility requirements
https://www.notion.so/bitliege/Skill-Accessibility-Requirements-42a7bd9f17a04c3eb8ee333848ef454b

Order of priority:
- Aria labels, labeling actions
- cleaning up how focus works
- arrow key support
- removing right pane from tab loop (?)

yonikid15:
Resolves #5249 

### Notes:
- Adds [accessibility improvements](https://www.notion.so/Skill-Accessibility-Requirements-42a7bd9f17a04c3eb8ee333848ef454b) to the find skills dialog. 
- Tabbing backward and forward between the tabable elements, while within a subcategory, still needs to be finished up. It's acting a bit strange in areas.
- This PR includes refactoring the Dialog and Accordion H2 components to work without scripts. (#5463)
- The `FindSkillsModal` component was getting a bit unruly so I split it up into 3 files in a new folder called `FindSkillsDialog`. It splits up the two sections into there own components: `SkillCategories.tsx` and `SeachResults.tsx`.
- Also, there are a few new messages that will need review and translations.


resources/assets/js/components/FindSkillsDialog/FindSkillsDialog.tsx
```diff
@@ -0,0 +1,435 @@
...
+  } = parentSkillCategories.reduce(
+    (
+      collection: { [key: string]: boolean },
+      parentSkillCategory: SkillCategory,
+    ) => {
+      collection[parentSkillCategory.key] = false;
+      return collection;
+    },
+    {},
+  );
```
tristan-orourke: skillsAccordionData and categoriesAccordionData are calculated every time the component renders, but are only used once, the first time the accordionData state hook is initialized. I think we should move the reduce calculations directly into the useState initialization parameter. We can compensate for it being less readable by adding some comments.

tristan-orourke: Using the queries.mapToObjectTrans utility function would probably help keep it readable.

resources/assets/js/components/FindSkillsDialog/FindSkillsDialog.tsx
```diff
@@ -0,0 +1,435 @@
...
+  // List of skills that displayed in the results section of the modal.
+  const [results, setResults] = useState<Skill[]>([]);
+  // Stores the skill category's name and description for the results section.
+  const [resultsSectionText, setResultsSectionText] = useState<{
+    title: string;
+    description: string;
+  }>({ title: "", description: "" });
+  // This state controls the search results panes display.
+  // If true it will display the default state, otherwise it will show the search results.
+  const [firstVisit, setFirstVisit] = useState(true);
```
tristan-orourke: We set this to false and back to true all the time, which means firstVisit isn't quite an accurate name :P

resources/assets/js/components/FindSkillsDialog/FindSkillsDialog.tsx
```diff
@@ -0,0 +1,435 @@
...
+    (event) => {
+      if (event.key === "Tab" && dialogRef.current) {
+        const tabableElements = Array.from(
+          dialogRef.current.querySelectorAll(
+            "[data-tabable='true']:not([disabled])",
+          ),
+        ) as HTMLElement[];
+
+        if (tabableElements.length === 0) {
+          event.preventDefault(); // TODO: should this throw an error?
```
tristan-orourke: Let's decide properly what to do, and get rid of that TODO.
I'm thinking now that preventDefault is probably the worst solution - it stops the tab key from working, and it does so silently 😆 . I'm thinking we should either throw an error ("A valid dialog must contain at least one focusable element") or we return _without_ preventing default, which allows the browser to handle the tab behavior. @yonikid15 what do you think.

resources/assets/js/components/FindSkillsDialog/FindSkillsDialog.tsx
```diff
@@ -0,0 +1,435 @@
...
+          ? tabableElements[backwardIndex]
+          : tabableElements[forwardIndex];
+
+        targetElement.focus();
+        prevTabbedElement.current = targetElement;
+        closeAllAccordions();
+        event.preventDefault();
+      }
+    },
+    [prevTabbedElement],
```
tristan-orourke: Now that prevTabbedElement is a ref hook, it doesn't need to be included in the array that causes recalculation. closeAllAccordions, on the other hand, should be. But closeAllAccordions is defined every render, so it needs to be in a useCallback hook as well for the memoization of handleTabableElements to be worth it.

resources/assets/js/components/FindSkillsDialog/FindSkillsDialog.tsx
tristan-orourke: I ended up handling this change myself while testing stuff out.

resources/assets/js/components/FindSkillsDialog/SearchResults.tsx
```diff
@@ -0,0 +1,350 @@
...
+    }
+  }, [results]);
+
+  const resultsRef = React.useRef<HTMLDivElement | null>(null);
+
+  React.useEffect(() => {
+    const keyHandler = (e: KeyboardEvent) => {
+      if (e.key === "Tab") {
+        resetResults();
+        e.preventDefault();
```
tristan-orourke: I think we don't want to prevent default here? I think we still want the FindSkillsDialog tab handler to run after we resetResults.

resources/assets/js/components/FindSkillsDialog/SearchResults.tsx
```diff
@@ -0,0 +1,350 @@
...
+    id: string,
+    newSkill: Skill,
+    newSkillsState: Skill[],
+  ): void => {
+    if (resultsRef.current) {
+      const skillResultsTabList = getTabList(resultsRef.current);
+      const checkbox = document.getElementById(id) as HTMLInputElement;
+      switch (e.key) {
+        case "ArrowLeft":
+          resetResults();
```
tristan-orourke, line 122: Should we preventDefault after handling an arrow key? (Here and for backBtnKeyDown too.) While testing, I see that using up and down arrows also causes the background to scroll, which I don't think is intentional.

resources/assets/js/components/FindSkillsDialog/SearchResults.tsx
```diff
@@ -0,0 +1,350 @@
...
+                {results.map((skill, index) => {
+                  const { id } = skill;
+                  const isAdded = newSkills.find(
+                    (newSkill) => newSkill.id === skill.id,
+                  );
+                  const isPreviousSkill =
+                    previousSkills.find(
+                      (previousSkill) => previousSkill.id === skill.id,
+                    ) !== undefined;
+                  const resultId = `skill-${id}`;
```
tristan-orourke: Given that this is used as an HTML id, which has to be unique on the page, maybe this should be a little more specific? 'skill-accordion-${id}`? I know the id for accordionData in FindSkillsDialog would need to be changed to.

tristan-orourke: Actually, the fact that two different components need to know which arbitrary string pattern to use as the same key is a recipe for future bugs. Lets see if we can improve that.
Maybe updating the shape of accordionData would help? Maybe something like:
```
{
   skills: {[id: string]: boolean},
   parentSkillCategories: {[key: string]: boolean}
}
```
That makes it a lot easier to see how to reference the right boolean, just by looking at the type definition.

resources/assets/js/helpers/forms.ts
```diff
@@ -43,6 +50,56 @@ export const toggleAccordion = (elementId: string): void => {
...
+ * @returns
+ */
+export const getTabList = (parentElement: HTMLElement | null): HTMLElement[] =>
+  parentElement
+    ? (Array.from(
+        parentElement.querySelectorAll(
+          "button:not([disabled]), [href], input:not([disabled]), select, textarea, [tabindex]:not([tabindex='-1'])",
+        ),
+      ) as HTMLElement[])
+    : [];
```
tristan-orourke: This should now be consolidated with getFocusableElements, a similar function I wrote in parallel which has already been merged.

resources/assets/js/helpers/forms.ts
```diff
@@ -18,13 +18,20 @@ export const getFocusableElements = (
...
- * Focuses on a element with given id.
- * @param id
+ * Takes a selector or an HTMLElement and focuses on the element.
+ * @param x
  */
-export const focusOnElement = (elementId: string): void => {
-  const element = document.getElementById(elementId);
-  if (element) {
-    element.focus();
+export const focusOnElement = (x: HTMLElement | string): void => {
```
tristan-orourke: We now have a bunch of functions in this file that aren't _really_ about forms. I think we have enough of them now  to move them to their own helper file. Maybe called `focus`?

resources/assets/js/components/H2Components/Accordion.tsx
```diff
@@ -81,14 +105,23 @@ const Btn: React.FunctionComponent<BtnProps> = (props) => {
...
-const Content: React.FunctionComponent<GeneralProps> = (props) => {
-  useAccordionContext(); // Ensures sub-component can only be used within the Accordion component.
-  const { className, children, ...rest } = props;
+const Content: React.FunctionComponent<Div> = ({ children, ...rest }) => {
+  const { isExpanded } = useAccordionContext(); // Ensures sub-component can only be used within the Accordion component.
+  const contentRef = React.useRef<HTMLDivElement | null>(null);
+  React.useEffect(() => {
+    if (contentRef.current && isExpanded) {
+      const focusableElements = getTabList(contentRef.current);
+      if (focusableElements.length > 0) {
```
tristan-orourke, line 120: overrideFocusRules should probably apply here. And if that breaks our current FindSkillsModal... maybe we split it into two properties? overrideOpenFocusBehaviour and overrideCloseFocusBehaviour?

resources/assets/js/components/H2Components/Accordion.tsx
```diff
@@ -98,22 +131,41 @@ const Content: React.FunctionComponent<GeneralProps> = (props) => {
...
+    triggerPos,
+    overrideFocusRules,
+    children,
+    ...rest
+  } = props;
+  const [expanded, setExpanded] = React.useState<boolean>(isExpanded);
+  React.useEffect(() => {
+    // Need to add this useEffect hook to ensure the state changes when the Accordion acts a controlled component.
+    setExpanded(isExpanded);
+  }, [isExpanded]);
```
tristan-orourke, line 159: It is _really_ cool how you've set this up to work whether the parent wants to control its state or not. I want to look at this again in the future, because I think it might not be ideal for performance, but that can be for another PR.

resources/assets/js/components/H2Components/Dialog.tsx
```diff
@@ -1,14 +1,16 @@
...
-  h2DialogEnableOpenDialogs,
-} from "@hydrogen-design-system/system/dist/import/latest/components/dialog/scripts/dialog";
-import { GeneralProps, GeneralBtnProps } from "./utils";
-
-interface DialogContext extends GeneralProps {
+import { focusOnElement, getTabList } from "../../helpers/forms";
+import { GeneralBtnProps } from "./utils";
+
+type Div = React.HTMLAttributes<HTMLDivElement>;
+type Heading = React.HTMLAttributes<HTMLDivElement>;
```
tristan-orourke: Two identical constants seems unnecessary :P I also find the names just a bit confusing - could we call it DivProps instead?

resources/assets/js/components/H2Components/Dialog.tsx
```diff
@@ -20,101 +22,97 @@ const DialogContext = React.createContext<DialogContext | undefined>(undefined);
...
+type TriggerProps = GeneralBtnProps;
+const Trigger: React.FunctionComponent<TriggerProps> = ({
+  id,
+  buttonStyling,
+  children,
+  ...rest
+}) => {
   return (
     <button
+      aria-haspopup
```
tristan-orourke: Looks to me like aria-haspopup needs [needs a value](https://www.digitala11y.com/aria-haspopup-properties/) of "true" or "dialog".

resources/assets/js/components/H2Components/Dialog.tsx
```diff
@@ -123,39 +121,124 @@ const Trigger: React.FunctionComponent<TriggerProps> = (props) => {
...
+      }
+    }
+  }, [isVisible]);
+
+  const handleKeyUp = React.useCallback(
+    (event) => {
+      switch (event.key) {
+        case "Escape":
+          closeDialog();
+          focusOnElement(`[data-h2-dialog-trigger=${id}]`);
```
tristan-orourke, line 188: This isn't quite a full solution - because we could potentially have multiple triggers for one dialog, and it wouldn't necessarily return to the right one.

tristan-orourke: I'm still not sure how to handle this best.

tristan-orourke: But I would consider fixing this to be the purview of #5463

resources/assets/js/components/H2Components/Dialog.tsx
```diff
@@ -123,39 +121,124 @@ const Trigger: React.FunctionComponent<TriggerProps> = (props) => {
...
+          break;
+        case "Tab":
+          if (dialogRef && dialogRef.current) {
+            const focusableDialogElements = getTabList(dialogRef.current);
+
+            if (focusableDialogElements.length === 0) {
+              event.preventDefault(); // TODO: should this throw an error?
```
tristan-orourke: Again... probably yes?

resources/assets/js/components/H2Components/Dialog.tsx
```diff
@@ -123,39 +121,124 @@ const Trigger: React.FunctionComponent<TriggerProps> = (props) => {
...
+            if (!event.shiftKey && document.activeElement === lastElement) {
+              firstElement.focus();
+              event.preventDefault();
+              return;
+            }
+
+            if (event.shiftKey && document.activeElement === firstElement) {
+              lastElement.focus();
+              event.preventDefault();
+            }
```
tristan-orourke, line 236: Can/should we use the helper functions you wrote here?

resources/assets/js/stories/FindSkillsDialog.stories.tsx
```diff
@@ -0,0 +1,42 @@
...
+        />
+        <FindSkillsDialog
+          previousSkills={[fakeSkill2(), fakeSkill4()]}
+          portal="applicant"
+          skills={fakeSkills()}
+          skillCategories={fakeSkillCategories()}
+          handleSubmit={async (x) => {
+            await new Promise((resolve) => setTimeout(resolve, 1000));
+            action("Submit Skills")(x);
+          }}
```
tristan-orourke: Just so you know, I wrote a shortcut helper function for this kind of thing at `resources\assets\js\stories\helpers.ts`

resources/assets/js/stories/FindSkillsDialog.stories.tsx
```diff
@@ -0,0 +1,42 @@
...
+import { withIntl } from "storybook-addon-intl";
+import { action } from "@storybook/addon-actions";
+import { useState } from "@storybook/addons";
+import { fakeSkill2, fakeSkill4, fakeSkills } from "../fakeData/fakeSkills";
+import { fakeSkillCategories } from "../fakeData/fakeSkillCategories";
+import FindSkillsDialog, {
+  FindSkillsDialogTrigger,
+} from "../components/FindSkillsDialog/FindSkillsDialog";
+
+const stories = storiesOf("Components/Find Skills Dialog", module).addDecorator(
```
tristan-orourke: Can we move this into "Applicant Profile/Skills"? I always forget where this Story is in the menu 😅

resources/assets/sass/h2.scss
```diff
@@ -8,3 +8,10 @@
   $hover-bg-color: #5ea8e4
 );
 @use "../../../node_modules/@hydrogen-design-system/system/dist/import/latest/system.scss";
+
+// TODO: Move this to new file?
```
tristan-orourke: 🤷

tristan-orourke (changes requested):

resources/assets/js/components/FindSkillsDialog/FindSkillsDialog.tsx
yonikid15: Well technically a dialog should **always** have at least a close button, so it shouldn't ever enter this block. I'm fine with with removing the prevent default, but then we should add in a close button in the top right corner (ie. x icon) that's permanently on every dialog.

resources/assets/js/components/H2Components/Dialog.tsx
yonikid15: I don't really see any improvements using the helper methods here, what do you think?

```js
if (!event.shiftKey) {
  focusPreviousItem(tabableElements);
  event.preventDefault();
  return;
}
```

resources/assets/js/components/FindSkillsDialog/FindSkillsDialog.tsx
tristan-orourke: Can we guarantee that, with the way we allow end users to build the contents of the modal out of sub-components?

yonikid15: We can permanently add it to the `<Header>` sub component, leave it trailing after the children.

resources/assets/js/components/H2Components/Dialog.tsx
tristan-orourke: Yeah, you're right. Your current solution allows the default behaviour to happen in most cases, which is probably for the best.

resources/assets/js/components/H2Components/Dialog.tsx
yonikid15: That is a mistake haha :p

resources/assets/js/components/H2Components/Dialog.tsx
yonikid15: Good point, not sure atm either.

tristan-orourke (approved):

PR #5508, opened 2021-03-25
Bugfix - application timeline skills & fit step save and quit function

### LINKED ISSUE #5352: Bug - Application Timeline: Skills - Save and quit option not working as intended
# Description
On the skills step in the application timeline, the applicant is prevented from saving their work and quitting to the my applications page. The form forces the user to first fill out all skill descriptions before leaving. We need to ensure that when the user clicks Save and Quit that it saves the current data to the backend, and then redirects them back to the My Applications page.

The same behaviour occurs in the fit step as well.

# Steps to Reproduce
1. Go to the skills step of the application timeline.
2. Click on save and quit (with missing information on at least 1 skill)
3. See error

yonikid15:
Resolves #5352 

### Notes:
- Ensures user can save and quit skills step without forcing them to fill out all the justifications.
- If they select "Save and Quit" and none of the forms have been filled it will simply jump to applications page.
- If they select "Save and Quit" and they have changed some or all of the forms. It will save all the forms that pass validation (not empty) and then go to the applications page. 
- One thing that may or may not be a problem is since you are leaving the page without saving it triggers `<AlertWhenSaved>`, which causes the confirmation alert to popup.
- Update: Also fixes the Fit steps save and quit behaviour as well.


tristan-orourke (changes requested):
If you press save and quit, and you have one or more invalid (ie empty) forms, none of them will be saved. I think you still want the valid ones to be saved?

yonikid15:
> If you press save and quit, and you have one or more invalid (ie empty) forms, none of them will be saved. I think you still want the valid ones to be saved?

Nice catch, currently trying to fix this now. It's a little trickier then I thought.

resources/assets/js/components/Application/Skills/Skills.tsx
```diff
@@ -8,9 +8,10 @@ import React, {
   RefObject,
 } from "react";
 import { FormattedMessage, useIntl, IntlShape } from "react-intl";
-import { Formik, Form, FastField, FormikProps } from "formik";
+import { Formik, Form, FastField, FormikProps, FormikValues } from "formik";
 import * as Yup from "yup";
 import Swal, { SweetAlertResult } from "sweetalert2";
+import { isEmpty } from "lodash";
```
tristan-orourke: Remember, since lodash is a big library it allows you to import more efficiently by doing `import isEmpty from "lodash/isEmpty"`

tristan-orourke: Actually its not being used, I'll just delete it :P

tristan-orourke (approved):
The way the "Are you sure you want to leave?" dialog appears even if every _valid_ form has been saved is a little annoying, but its not _entirely_ unreasonable, so whatever. Your changes look good!

PR #5509, opened 2021-03-25
Fix assorted typos

mnigh:
Fixes a range of typos in comments, copy, keys, and variable names.

### Notes:
- Identified typos with help from [Code Spell Checker linter for Visual Studio Code](https://marketplace.visualstudio.com/items?itemName=streetsidesoftware.code-spell-checker).


resources/assets/js/models/localizedConstants.tsx
```diff
@@ -465,63 +465,63 @@ const provinceNames = defineMessages({
 export const provinceName = (provinceId: number): MessageDescriptor =>
   getOrThrowError(provinceNames, provinceId, "invalid ProvinceId");
```
yonikid15, line 466: I see there are two functions for getting the province abbreviations: `provinceName` and `provinceAbbreviations`. We should remove one and make sure that the correct one is getting referenced throughout the app (right now `provinceName is getting referenced).

database/migrations/2018_12_21_175437_populate_skills.php
```diff
@@ -63,7 +63,7 @@ public function up()
             ['skill_type_id' => 2, 'name' => 'project_management'],
             ['skill_type_id' => 2, 'name' => 'stakeholder_relations'],
             ['skill_type_id' => 2, 'name' => 'dot_net'],
-            ['skill_type_id' => 2, 'name' => 'geospacial_programming'],
+            ['skill_type_id' => 2, 'name' => 'geospatial_programming'],
```
yonikid15, line 66: You shouldn't ever go back to an old migration and make changes, but instead create a new to make whatever changes you need. However, I think this case might be fine since it's only fixing a typo on the name field and not manipulating the structure or a value that is referenced anywhere. Let me double check and get back to you though.

yonikid15 (changes requested):

database/migrations/2018_12_21_175437_populate_skills.php
yonikid15: So after talking to Tristan, you'll need to undo this change and instead create a new migration that fixes the spelling. This is due to the fact that it won't actually change the value in production, but make it seem like it did. Plus, you also fixed the spelling in the lang files which are getting referenced by the DB, so this will break the connnection.

mnigh: Yep, I recognized the fact that I was trying to manipulate an existing migration, but I'm pretty sure we came to the same conclusion: that the change does not change the structure or value of anything referenced in the codebase (the skills table's structure has been changed since the aforementioned migration, including changing the name column from a string to json type). Do let me know if this still poses a problem though, thanks.

yonikid15: After further investigation this seems fine.

resources/assets/js/models/localizedConstants.tsx
yonikid15: Sorry I'm wrong, that's a different list of messages. This is correct.

yonikid15 (approved):

PR #5594, opened 2021-04-06
Test php-actions/composer

mnigh:
### Notes:
- Testing value for [existing issue](https://github.com/php-actions/composer/issues/53#issuecomment-814194946) with php-actions/composer GitHub action.


g105b:
For this test to succeed you'll need to [enable the improved container support](https://docs.github.com/packages/getting-started-with-github-container-registry/enabling-improved-container-support).

